PR #6, opened 2018-11-06
Ansible playbooks

pietroalbini:
Description of most the choices:

https://paper.dropbox.com/doc/Ansible-setup-for-Rust-infra--ARLTbCd_DJNnuhf0pNbLWeF2Ag-Hie5adciBykpNjZqYQFb3


ansible/ansible.cfg
```diff
@@ -0,0 +1,3 @@
+[defaults]
```
alexcrichton: I'm personally not very familiar with ansible, so this file seems entirely magical to me both in existence, what the keys are, and what they're configured to. Could files perhaps have a link to a URL describing them and/or relevant documentation?

ansible/deploy
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
alexcrichton: I personally find bash scripts to be pretty awful for long-term usage, especially when they're managed by multiple folks (everyone seems to always have their own style). This is pretty small though, so would something like Python perhaps work for the script?

ansible/group_vars/all.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+# Use python3 on the remote servers
+# Ubuntu doesn't ship with python3 preinstalled anymore.
```
alexcrichton: Speaking as an outsider at least, this comment doesn't really make much sense to me. It looks like we're configuring to use an interpreter that isn't installed in various places? Not that I really know what's going on with this file anyway...

pietroalbini: Uh, woops, the comment is wrong, it should be "Ubuntu doesn't ship with python2 preinstalled anymore". Ansible by default uses Python 2, but it can be configured to run with Python 3.

PR #15, opened 2019-09-10
Revoke misdreavus's old work ssh key

QuietMisdreavus:
My work computer has its own set of SSH keys, but since i'm going to be leaving at the end of this week, it's worth cleaning up to make sure i don't leave any extra keys behind.

I'm posting this today, but please *do not deploy* this until this Saturday, 2019-09-14. I still have use of the computer this week, and i want to be able to respond to docs.rs incidents if necessary. Thanks!


pietroalbini:
Will the whitelisted IP also need to change?

QuietMisdreavus:
No. The whitelisted IP is for my private server, which will still be accessed from my other machines.

pietroalbini:
Deployed.

PR #18, opened 2019-10-17
Fix "typo"

gnzlbg:


pietroalbini:
Well, "validare" actually correct... in Italian :sweat_smile: 

Thanks for spotting this!

gnzlbg:
> Well, "validare" actually correct... in Italian 😅

That's why I put it in quotes 😆

PR #21, opened 2020-01-08
Security / Public PEM Key

bryantbiggs:
- Remove public pem key string contents and replace with SSM parameter data resource for lookup at deploy time.

Note: This change requires someone with access to the AWS environment to create a SSM secure string parmater named `/buildabot/ec2_key/public` with the public key contents

### DIFF
--- a/terraform/ec2.tf
+++ b/terraform/ec2.tf
@@ -13,9 +13,14 @@ data "aws_ami" "ubuntu_bionic" {
   }
 }
 
+data "aws_ssm_parameter" "buildbot_west_slave_key" {
+  name            = "/buildabot/ec2_key/public"
+  with_decryption = true
+}
+
 resource "aws_key_pair" "buildbot_west_slave_key" {
   key_name   = "buildbot-west-slave-key"
-  public_key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCdGoRV9XPamZwqCMr4uk1oHWPnknzwOOSjuRBnu++WRkn7TtCM4ndDfqtKnvzlX5mzPhdvO1KKx1K8TiJ3wiq7WS4AFLGKQmPHWjg8qxGW7x4S8DHrb4ctmaujZ1+XCNSK3nsCl1lLW8DOrRlKbfeHIAllbMBZxIRmQ+XICVvhKAmSmxzTmYC8tBqvqQprG/uIuKonjLxL/ljtBxXBNECXl/JFCYG0AsB0aiuiMVeHLVzMiEppQ7YP/5Ml1Rpmn6h0dDzFtoD7xenroS98BIQF5kQWhakHbtWcNMz7DVFghWgi9wYr0gtoIshhqWYorC4yJq6HGXd0qdNHuLWNz39h buildbot-west-slave-key.pem"
+  public_key = data.aws_ssm_parameter.buildbot_west_slave_key.value
 }
 
 resource "aws_security_group" "legacy_common" {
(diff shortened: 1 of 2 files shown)


pietroalbini:
Thanks for the PR!

Is there a reason why you suggest putting the public SSH key in Parameter Store? Unless something is sensitive (such as whitelisted IP addresses for the security groups, and actual private tokens) we try to have everything as public as possible in our infrastructure.

bryantbiggs:
Its a manually generated resource thats been plugged in by hand. If you need to rotate keys, that requires more manual intervention. Here is a pattern I use often for this exact scenario:

```hcl
resource "tls_private_key" "ec2_pem" {
  algorithm = "RSA"
  rsa_bits  = "4096"
}

resource "aws_key_pair" "ec2_pem" {
  key_name   = "some_key_pair"
  public_key = tls_private_key.ec2_pem.public_key_pem
}

resource "aws_ssm_parameter" "ec2_pem" {
  name        = "some_key_pair"
  description = "EC2 private PEM key"
  type        = "SecureString"
  value       = tls_private_key.ec2_pem.private_key_pem
}
```

No manually entered values, and users can access the private key portion from SSM. Also, if this is for SSH access into the bastion - that should be dropped all together in favor of using SSM session manager. This removes the need for SSH key management all together and is more secure. SSH keys in AWS should only be required for external parties requiring a persistent SSL proxy through a bastion host - typically something like an external business intelligence tool that is making a JDBC/ODBC connection to a database in the account.

pietroalbini:
That snippet to generate the SSH key on the fly with Terraform is cool, and I definitely see the value of it. I see less value though in just moving the public key to Parameter Store: there is no automatic rotation setup, so we'd still have to document to change the key in Parameter Store instead of this repo.

We haven't yet decided as a team how much we want to commit to Terraform: if we do that I'll probably use something like that snippet to generate the key in Terraform, otherwise I'll just be here on the repo.

> Also, if this is for SSH access into the bastion - that should be dropped all together in favor of using SSM session manager.

The key is not used to access the bastion, nor any other system on a regular basis: the only purpose for it is to apply the initial configuration with Ansible, and once that's done everyone uses their own keys to access the systems they're authorized to log into. 

We don't really want to use Session Manager: we have other machines outside of AWS, and having a single way to authenticate to all our instances across providers is valuable for us.

bryantbiggs:
> We don't really want to use Session Manager: we have other machines outside of AWS, and having a single way to authenticate to all our instances across providers is valuable for us.

I think that is because you haven't tried it or are unfamiliar with it and the concept. What I am trying to say is that the way you currently approach infrastructure management is not the way its performed per the best practices of today. Dynamic configuration management with Ansible is going away in favor of immutable infrastructure. I can help with this - its common to use the ansible provider with Packer to build an AMI but after that artifact is generated, its not dynamically modified until a new image needs to be created (usually when updating software, there is a separate process for patching vulnerabilities). 

With SSM session manager - you already authenticate against AWS no matter what, most likely through your `~/aws/credentials` file. This is unique for each cloud provider. Then instead of `ssh user@host.com -i pem` you would do `aws ssm start-session --target 7.8.9.10` and your IAM creds are used to authenticate and check authorization before making the remote connection. Authorization is set in IAM - more secure than physically placing a key on the host, less overhead and management, fast access grant and revocation. 

And regarding Terraform - I'm not sure what alternative the infra team would go with. Of the cloud agnostic IAC tools - terraform and pulumi, terraform has the clear lead and advantage. If there is hesitation around terraform then I would suspect that is due to the lack of experience and hence why I am reaching out to offer that assistance.

Issue #22, opened 2020-01-08
Terraform state management


bryantbiggs:
It is a recommended best practice to segment your terraform state so that you can limit the blast radius in the event of adverse modifications. How it is segmented is completely subjective, but I would suggest segmenting it based on service/functionality. 

For example:
- `docs-rs`
- `crates-io`
- `rust-lang`
- `highfive`
- `global` - things like Route53 zones, general IAM settings not specific to above services, etc.

This is just a rough first pass but it demonstrates the approach. Changes to `docs-rs` will not take down or affect any other services when changes to the infrastructure are made. It also speeds up the develop/deploy process as less resources are managed under each statefile and less time is traversing the statefile to compare and report.

Reach out to me at bryantbiggs@clowd.haus and lets chat about overall infrastructure. I have been working with Terraform for over 4 years now and have built out/transformed several environments over the years and am willing to help here if there is interest.

pietroalbini:
Yeah, this is a good point, and splitting the state file was one of the things I wanted to do next.

This setup started as a sort of exploration on how suitable Terraform was for us: we're an open source project with an infra team made mostly of volunteers, so our requirements are a bit different than traditional infrastructure (we still need to decide if and how much we want to commit to Terraform in the long term).

Unfortunately a few of the early decisions (such as having a single state file to reduce complexity) were not great, and some of the initial configuration could be rewritten in a better way, but if we commit to Terraform in some way we'll definitely spend time cleaning them up.

> Reach out to me at bryantbiggs@clowd.haus and lets chat about overall infrastructure. I have been working with Terraform for over 4 years now and have built out/transformed several environments over the years and am willing to help here if there is interest.

Having a chat would be great! I'd slightly prefer to do it [on our Discord](https://discord.gg/rust-lang) (channel `#infra`) instead of email, to allow other interested members of the team to chime in. Let me know if that's a problem for you.

PR #23, opened 2020-01-09
Chore / Local Variable Clean-Up

bryantbiggs:
- Use local variables to consolidate DNS names under single variable
- Merge `redirects.tf` and `route53.tf` under file `dns.tf`


pietroalbini:
Thanks for the PR!

> Use local variables to consolidate DNS names under single variable

What's the reasoning behind this change?

> Merge redirects.tf and route53.tf under file dns.tf

I'd prefer not to do that: `route53.tf` is supposed to eventually go away, while we migrate all domains to `services/dns/{domainname}.tf`. Once we do that, having just redirects in `dns.tf` would be a bit confusing.

bryantbiggs:
Its a good practice, in all facets of software, to use variables over copying+pasting the same value. 

Regarding the `services/dns/{domainname}.tf` - I would advise against this. Its recommended by Hashicorp to keep your structure as flat as possible - avoiding the usage of modules within modules. This is where I started to go with #22 but its much easier to discuss rather than type out. There are quite a few areas of improvement in the overall management of cloud infrastructure based on what I see in the repository and thought a conversation would be more useful that GitHub issues/PRs

pietroalbini:
> Its a good practice, in all facets of software, to use variables over copying+pasting the same value.

That's for sure! I just see little value in adding variables for `"crates.io"` and `"rust-lang.org"`. Most of the Terraform configuration already uses variables for the domain names, the only two places where domains are present are:

* For `services.tf` the repetition is not too bothersome to me, as that just defines the actual domains where those services will live.
* I agree with you that there is unnecessary repetition in the DNS records definition (`services/dns/{domainname}.tf`), but it's possible to just avoid the problem by changing the `service/dns/domain/` module to stop requiring every record to be prefixed with the domain name. I'll explore if that's possible.

> Regarding the services/dns/{domainname}.tf - I would advise against this. Its recommended by Hashicorp to keep your structure as flat as possible - avoiding the usage of modules within modules. This is where I started to go with #22 but its much easier to discuss rather than type out.

For this I'll probably split out `services/dns` into its own state file (or even, all the subdirectories of `service/`), with #22.

PR #25, opened 2020-01-22
Use rust base64 instead of os dependent one

Zexbe:
Base64 isn't the same on all distributions, so to make it more cross-platform, we can use the rust crate base64

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,6 +75,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


pietroalbini:
Thanks for the PR!

While I agree that using the base64 crate in `src/main.rs` is desiderable, we should keep the old approach for `src/deploy.rs`: that script is [compiled directly with `rustc`](https://github.com/rust-lang/simpleinfra/blob/ac7e21a83fcb411e6e9fb9c5fb0dc10780f79176/travis-configs/static-websites.yml#L33), and we don't have an easy way to compile dependencies that way.

Zexbe:
@pietroalbini Could the line be changed to use cargo instead?

pietroalbini:
@Zexbe not really, that snippet is also present across other configuration files across the org, and we want the script to have as little overhead as possible (compared to fetching the index, downloading dependencies, building and linking them...).

I don't think it's a problem if we shell out in `deploy.rs`, as the environment the program is run on doesn't change over time (it's always a Linux builder on either Travis CI or GitHub Actions).

Zexbe:
@pietroalbini Okay, it's done

pietroalbini:
Thanks!

PR #26, opened 2020-01-23
Add run-github-release action.

XAMPPRocky:
This started out as a way to easily use mdbook for deploying, but it was easy enough to make it configurable to work with any github project's releases. I've tested that it works locally but I haven't tested it as a github action yet.


XAMPPRocky:
@pietroalbini I think there's a bug in CI, since I'm on macOS `ncc` produces a slightly different build than when on CI.

pietroalbini:
Could you try replacing `requests` with [`node-fetch`](https://www.npmjs.com/package/node-fetch)? It has less code and no dependencies, but more importantly it doesn't have the dependency causing the diff (`though-cookie`).

XAMPPRocky:
@pietroalbini I've migrated to `node-fetch`, but the diff persists.

github-actions/run-github-release/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    required: true
+  repo:
+    description: The name of the repository containing the releases.
+    required: true
+  token:
+    description: GitHub token for authenication.
+    required: true
+
+runs:
+  using: node12
```
meysam81, line 26: What is stopping you from migrating to a higher node version? This version will show a deprecation warning, and [GitHub will soon drop support](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

github-actions/run-github-release/README.md
```diff
@@ -0,0 +1,42 @@
...
+The `run-github-release` action allows you to run any binary release on GitHub.
+See [`actions.yml`] for configuration options.
+
+[`actions.yml`]: ./actions.yml
+
+## Example
+This example builds an mdbook project on linux.
+
+```yaml
+- uses: rust-lang/simpleinfra/github-actions/run-github-release@master
```
meysam81, line 12: It's probably better to pin this to a specific version so that future modifications won't break your existing users:

```suggestion
- uses: rust-lang/simpleinfra/github-actions/run-github-release@v1
```

Of course, there will be a need to create a new release in this repository, but it's better to be safe than sorry!

PR #28, opened 2020-02-10
Add simple-ci action

XAMPPRocky:
This Pr adds the "simple CI" github action. The intent with this action is to use it for projects in the Rust org where the CI script is often just `cargo build && cargo test`. With this action we can add configuration and functionality that can be shared across repositories.


pietroalbini:
Thanks!

Issue #32, opened 2020-03-13
Implement tooling to rollback a ECR service deployment


pietroalbini:
Currently, the way to rollback a service deployed to ECR is to push a new image and redeploy it, which can be unnecessarily slow.

ECR has support for [retagging an image without uploading it again](https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-retag.html). We should develop a script that prompts which image the user wants to rollback, retags it as `latest` and forces a new ECS deployment.

pepoviola:
Hi @pietroalbini, I can help with this script with some guidance.

pietroalbini:
That's great! My idea for this is to create a small Python script (like `aws-creds.py` in the root of this repository) that:

1. Calls the ECR API to list the available images in the repository.
2. Asks for the image the user wants to roll back to.
3. Tags the image the user chose to `latest`, without downloading it.
4. Forces a new ECS deployment.

Note that you'll need an AWS account of your own to test this, and I'm not sure if everything is included in the free tier.

pietroalbini:
@pepoviola are you still working on this?

pepoviola:
Hi @pietroalbini, sorry about the delay. It's ok to use the [python sdk](https://aws.amazon.com/sdk-for-python/) to interact with aws?

Thx!

pietroalbini:
@pepoviola I'd slightly prefer to shell out to awscli, to avoid potential issues with Python dependencies (the CLI outputs JSON anyway).

pepoviola:
Hi @pietroalbini, I have the logic in place for:

1. Calls the ECR API to list the available images in the repository.
2. Asks for the image the user wants to roll back to.
3. Tags the image the user chose to latest, without downloading it.

How you are forcing `ecs` to re-deploy, creating a new task definition?

Thanks!

pietroalbini:
@pepoviola we use this command:

```
aws ecs update-service --cluster rust-ecs-prod --service <service-name> --force-new-deployment
```

pepoviola:
Great, thanks @pietroalbini. I will create a pr for review.

pepoviola:
Hi @pietroalbini, can we close this issue?
Thx!

pietroalbini:
Yep! Thanks!

PR #34, opened 2020-04-10
Add rollback tool

pepoviola:
Hi @pietroalbini, this pr add a small python script for rollback ( re-tagging and force re-deploy ) related to  #32 .

A couple of `TODO`s: set the default values for 
- DEFAULT_REPOSITORY_NAME
- ECS_CLUSTER
- ECS_SERVICE

If you can we can support passing those as arguments.

Thanks!


aws-rollback.py
```diff
@@ -0,0 +1,138 @@
...
+import sys
+
+DEFAULT_REPOSITORY_NAME = "rust-central-station"
+TARGET_TAG = 'latest'
+ECS_CLUSTER = 'rust-ecs-prod'
+ECS_SERVICE = 'TODO COMPLETE'
+
+def main():
+    """CLI entrypoint of the program."""
+    repository_name = DEFAULT_REPOSITORY_NAME if len(sys.argv) < 2 else sys.argv[1]
```
pietroalbini, line 14: I'd prefer not to have a default repository name.

aws-rollback.py
```diff
@@ -0,0 +1,138 @@
...
+    images = get_images(repository_name)
+    while True:
+        selected_image_index = let_user_pick_image(images)
+        if selected_image_index == -1:
+            exit(0)
+
+        if not ( selected_image_index is None ):
```
pietroalbini, line 21:
```suggestion
        if selected_image_index is not None:
```

aws-rollback.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
+
+import json
+import subprocess
+import sys
+
+DEFAULT_REPOSITORY_NAME = "rust-central-station"
+TARGET_TAG = 'latest'
+ECS_CLUSTER = 'rust-ecs-prod'
+ECS_SERVICE = 'TODO COMPLETE'
```
pietroalbini, line 10: A solution here could be to list the services inside the cluster with:

```
aws ecs list-services --cluster {ECS_CLUSTER}
```

If a service with the same name as the repository exists we redeploy that service, otherwise we rollback but *not* redeploy.

aws-rollback.py
pepoviola: Great, just to clarify the `repository name` should be a required argument.

aws-rollback.py
pepoviola: Sounds good, and it ok to have the `ECS_CLUSTER` as a constant or you want to receive as argument?
Thx!

pepoviola:
HI @pietroalbini, thanks for the feedback! I just make a quick questions to review the changes.
Thanks!

aws-rollback.py
pietroalbini: Yep, we only have one cluster at the moment.

aws-rollback.py
pietroalbini: Yes.

pepoviola:
Hi @pietroalbini, I updated the code with the suggested changes.
Thanks!

aws-rollback.py
```diff
@@ -0,0 +1,164 @@
...
+    eprint(f"image {image} retaged as '{TARGET_TAG}'")
+    if can_redeploy(repository_name):
+        redeployed = force_redeploy()
+        if redeployed:
+            eprint(f"successfully rollback and re-deploy")
+
+def let_user_pick_image(images):
+    print("Please choosean image to rollback:")
+    for idx, image in enumerate(images):
+        print("{}) {}".format(idx+1,image["imageTag"]))
```
pietroalbini, line 40: In our repository, only the `latest` image is tagged, while all the past ones are untagged. It would be nice if the prompt looked something like this (the list is sorted from the most recently pushed image to the oldest):

```
Please choose an image to rollback:
2) 2020-04-13 12:02:39 (latest)
3) 2020-04-11 00:25:25
4) 2020-04-06 12:02:30 (other-tag)
5) 2020-04-01 13:15:24
```

You can use the `aws ecr describe-images` command instead of `aws ecr list-images` to fetch the additional information. Also, since not all the images are tagged you'll need to use digests to identify images.

aws-rollback.py
```diff
@@ -0,0 +1,164 @@
...
+    manifest = get_image_manifest(repository_name, image["imageTag"])
+    retag_image(repository_name,manifest)
+    eprint(f"image {image} retaged as '{TARGET_TAG}'")
+    if can_redeploy(repository_name):
+        redeployed = force_redeploy()
+        if redeployed:
+            eprint(f"successfully rollback and re-deploy")
+
+def let_user_pick_image(images):
+    print("Please choosean image to rollback:")
```
pietroalbini, line 38:
```suggestion
    print("Please choose an image to rollback:")
```

Small typo :)

aws-rollback.py
```diff
@@ -0,0 +1,164 @@
...
+        err(f"selected image already tagged as {TARGET_TAG}")
+
+    eprint(f"selected option: {image}")
+    manifest = get_image_manifest(repository_name, image["imageTag"])
+    retag_image(repository_name,manifest)
+    eprint(f"image {image} retaged as '{TARGET_TAG}'")
+    if can_redeploy(repository_name):
+        redeployed = force_redeploy()
+        if redeployed:
+            eprint(f"successfully rollback and re-deploy")
```
pietroalbini, line 35: It'd be nice here to show success message even for other cases:

* Successfully rolled back the image, but no service with the same name to redeploy found
* Successfully rolled back the image, but redeploying the service with the same name failed

aws-rollback.py
pepoviola: thanks! :)

aws-rollback.py
pepoviola: Hi @pietroalbini, thanks for the feedback. I just pushed all the changes.
Thx!

aws-rollback.py
```diff
@@ -0,0 +1,177 @@
...
+    try:
+        out = json.loads( run_command([
+            "aws", "ecr", "describe-images",
+            "--repository-name", repository_name,
+            "--no-paginate"
+        ]).stdout)
+    except subprocess.CalledProcessError as e:
+        err(f"failed to get availabe images from repository: {e}" )
+
+    return out["imageDetails"]
```
pietroalbini, line 76:
```suggestion
    return list(sorted(out["imageDetails"], key=lambda image: image["imagePushedAt"], reverse=True))
```

PR #35, opened 2020-05-15
simple-ci: fix cargo fmt invocation.

smarnach:
I tried to enable the format check, but it [broke the build](https://github.com/rust-lang/crates-io-ops-bot/runs/678093441?check_suite_focus=true). The `--check` flag is an argument to `rustfmt`, so we need to pass it after `--`.


smarnach:
Thanks for the quick turn-around! Unfortunately, this change [wasn't enough to fix the feature](https://github.com/rust-lang/crates-io-ops-bot/runs/678570559). I'll take another look tomorrow.

pepoviola:
Hi @smarnach, I think you need to add the `check_fmt` to check ( and install fmt ).

```
  with:
    check_fmt: true
```

I open a [pr](https://github.com/rust-lang/crates-io-ops-bot/pull/36) for fix this.
Thx!

smarnach:
@pepoviola I filed a fix in #36.

pepoviola:
HI @smarnach, thanks for the update 👍 .

PR #36, opened 2020-05-15
simple-ci: Fix the installation of the rustfmt component.

smarnach:
This is a follow-up for #35.

The old version of the check was incorrect. The check `which rustfmt` can succeed even when the `rustfmt` component is not installed for the current Rust toolchain. `cargo fmt`, on the other hand, only works when the component is installed. The mere presence of the `rustfmt` binary isn't enough.

The fix is to simply install the component unconditionally. If it is already installed, this is a no-op anyway.


XAMPPRocky:
This looks fine to me, we could also remove the `rustup component add rustfmt` since that is now shipped by default.

smarnach:
The reason I filed this PR is [that it wasn't working without installing the component](https://github.com/rust-lang/crates-io-ops-bot/pull/35/checks?check_run_id=678570559):

```
/usr/bin/which rustfmt
/usr/share/rust/.cargo/bin/rustfmt
/usr/share/rust/.cargo/bin/cargo fmt -- --check
error: 'cargo-fmt' is not installed for the toolchain 'nightly-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt`
```

This error is for the nightly build. So it looks like we can't just remove it.

XAMPPRocky:
Huh, that's odd.  Either way +1

PR #37, opened 2020-05-21
Add WG_AND_TEAMS env var

technetos:
This PR adds the WG_AND_TEAMS env var to the discords-mod-bot deployment environment so that we can limit certain commands to members of that role.


pietroalbini:
Thanks!

pietroalbini:
Deployed!

PR #43, opened 2020-07-14
Fix the Authorization header not being forwarded on crates.io

pietroalbini:
r? @jtgeibel

### DIFF
--- a/terraform/shared/services/cratesio/cloudfront-webapp.tf
+++ b/terraform/shared/services/cratesio/cloudfront-webapp.tf
@@ -41,6 +41,9 @@ resource "aws_cloudfront_distribution" "webapp" {
         // may want to forward it along at this layer as well. This might be
         // helpful for debugging at some point.
         "X-Request-Id",
+        // Some authenticated API endpoints use the GET method, so we need to
+        // forward the Authorization header to allow token authentication.
+        "Authorization",
       ]
       query_string = true
       cookies {


Turbo87 (approved):

jtgeibel:
I'm not sure we want to do this yet. We should at least review our current APIs first and restrict the endpoints that accidentally allow tokens instead of just cookies. I don't want to get locked in to allowing tokens on endpoints where we have accidentally allowed them.

jakeswenson:
I would love for this change to make it through. 
I am interested in building something to track my followed crates.


I did a quick query of all the uses of the `authenticate()` method

<details>
<summary>Results</summary>

```bash
 rg 'authenticate\(\)'
src/controllers/version/yank.rs
33:    let authenticated_user = req.authenticate()?;

src/controllers/crate_owner_invitation.rs
9:    let user_id = req.authenticate()?.user_id();
43:    let user_id = req.authenticate()?.user_id();

src/controllers/token.rs
12:    let authenticated_user = req.authenticate()?;
64:    let authenticated_user = req.authenticate()?;
100:    let authenticated_user = req.authenticate()?;

src/controllers/user/me.rs
17:    let user_id = req.authenticate()?.user_id();
58:    let authenticated_user = req.authenticate()?;
110:    let authenticated_user = req.authenticate()?;
195:    let authenticated_user = req.authenticate()?;
236:    let user_id = req.authenticate()?.user_id();

src/controllers/krate/owners.rs
87:    let authenticated_user = req.authenticate()?;

src/controllers/krate/search.rs
45:    let authenticated_user: AppResult<AuthenticatedUser> = req.authenticate();

src/controllers/krate/follow.rs
24:    let user_id = req.authenticate()?.user_id();
37:    let user_id = req.authenticate()?.user_id();
49:    let user_id = req.authenticate()?.user_id();

src/controllers/krate/publish.rs
62:    let ids = req.authenticate()?;
```

</details>

of those, the ones that are `GET`'s as defined in `router.rs` are:

- list invitations `/api/v1/me/crate_owner_invitations`
- list tokens `/api/v1/me/tokens` (of which great care was taken to avoid possibly exposing the token secret)
- get following status for a crate `/api/v1/crates/:crate_id/following`
- get user details `/api/v1/me` (returns `EncodablePrivateUser`)
- get following updates `/api/v1/me/updates` 
- get ("search") crates `/api/v1/crates` (only when `following` param is passed, to return the crates the user is following)

Please independently verify.

Are any of these APIs concerning to expose @jtgeibel?

jakeswenson (approved):

pietroalbini:
We'll need for rust-lang/crates.io#3470 to be deployed to production, once that's done we can apply and merge this!

jtgeibel:
rust-lang/crates.io#3470 is now deployed on https://staging.crates.io. I'd like to also land rust-lang/crates.io#3518 before landing this. In particular, the `/me` endpoint returns the users verified email address and I would prefer to keep that to cookie auth only.

jtgeibel:
@jakeswenson one quick question about `/api/v1/crates/:crate_id/following`. Do you plan to use that endpoint for anything? We currently use this in the frontend (via cookie auth), but I think it is kind of silly to make a request to get a single boolean value. I think it makes sense to restrict this endpoint as well, unless there is a specific use case for it.

jtgeibel:
rust-lang/crates.io#3518 is now deployed, enabling just `/api/v1/crates/:crate_id/following` and `/api/v1/crates?following=1`, which should unblock this.

jtgeibel (approved):

pietroalbini:
Applied! The change should be live within the next 15 minutes.

PR #44, opened 2020-07-24
updates config for the crates io bot to include new environmental var…

nellshamrell:
…iable.

Also updates a few other environmental variables to load their value from the AWS parameter store, rather than being hard coded.

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>


pietroalbini:
@nellshamrell I'd prefer for the non-secret variables to be hardcoded: one of the main reasons why we're switching to Terraform is to allow non-infra people to take a look at our infra and submitting PRs. Making the values public allows, for example, a crates.io team member to know what the value is and propose a tweak.

nellshamrell:
Understood, I'll modify the PR to do that.

nellshamrell:
Closing in favor of #45

PR #46, opened 2020-08-19
slight correction to default value for the volumes variable, needs to…

nellshamrell:
… be an empty array for a list rather than null


pietroalbini (approved):
Feel free to merge once CI passes

PR #49, opened 2020-08-20
Automatically forward requests through the bastion

pietroalbini:
Our databases are hosted with RDS on a private subnet, and are not reachable from the public internet. Instead, you have to connect to them through our bastion server, increasing the security of the data we store.

Unfortunately Terraform's PostgreSQL provider doesn't support any kind of proxy or port forwarding, but it only allows direct connections. To work around the problem, we call this script from an "external" Terraform provider: this script then setups the port forwarding using SSH, and returns the local address and port for Terraform to connect to.

This script only has one limitation: the "external" Terraform provider doesn't support calling a command to cleanup the resources once the Terraform command exits, so we're forced to leak the SSH connection. To work around that it's possible to configure a timeout, and the script will reuse existing connections whenever possible.

r? @Mark-Simulacrum (it'd be also good to know whether it works on your machine)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /ansible/.venv
 /ansible/envs/dev
 /terraform/shared/modules/lambda/packages
+/terraform/rds-databases/.forward-ports-cache-*.json
 .terraform
 node_modules
 __pycache__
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
How do I go about testing this? Just run `terraform plan` or something like that in rds-databases?

pietroalbini:
Yep!

Mark-Simulacrum:
Seems to work well. I was worried that it would break when the socket is already bound, but I think the script handles that today.

r=me

PR #50, opened 2020-08-21
Reduce level of access to our infrastructure

pietroalbini:
This PR reduces the level of access to our infrastructure:

* Now that the RDS instance is available from the bastion over the private network, and Terraform is configured to proxy through it, we don't need to grant access from the public internet to allowed IPs.
* Removed some IPs from the bastion security group, as those people don't have access to the bastion anymore.

r? @Mark-Simulacrum

### DIFF
--- a/terraform/bastion/firewall.tf
+++ b/terraform/bastion/firewall.tf
@@ -8,13 +8,8 @@ locals {
   //     /prod/bastion/allowed-ips/${user}
   //
   allowed_users = [
-    "acrichto",
     "aidanhs",
-    "guillaumegomez",
     "joshua",
-    "mozilla-mountain-view",
-    "mozilla-portland",
-    "mozilla-san-francisco",
     "onur",
     "pietro",
     "shep",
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum:
Before doing this, I would like us to have a script similar to aws-creds.py which permits psql connections to our shared database instance. I currently frequently need to check or adjust something for triagebot and rustc-perf, and having to set up ssh forwarding (remembering the various DNS names and such) is annoying.

pietroalbini:
Note that you can run `psql` on the bastion now. Still, I'll create a `aws-psql.py` script :)

Mark-Simulacrum:
Oh, if I can do that, that's enough for me.

Mark-Simulacrum:
Well -- kind of enough, the bastion should be updated to ubuntu 20.04 probably (the latest LTS), instead of 18.04:

```
psql (10.12 (Ubuntu 10.12-0ubuntu0.18.04.1), server 12.2)
WARNING: psql major version 10, server major version 12.
         Some psql features might not work.
```

pietroalbini:
Merging this, will publish the `aws-psql.py` script in a followup PR.

PR #61, opened 2020-12-30
Add `custom_domain` key for github-actions/static-websites

JohnTitor:
This would help migration to GHA on rustc-dev-guide.
r? @pietroalbini


pietroalbini:
Thanks for the PR!

I don't think this action should actually be used: being based on Docker it's actually pretty slow :(

JohnTitor:
So, we should just deploy files manually or is there any other way? I thought we could use this action for rust-lang/rustc-dev-guide#940.

PR #64, opened 2021-03-03
Avoid duplicate rules for rustc-builds and rustc-builds-alt

Mark-Simulacrum:
These rules overlapped (the prefix for both is rustc-builds) and their contents
were identical regardless, so this is pointless.

r? @pietroalbini


pietroalbini (approved):
Looks good to me!

You can apply the change, but i'd like a comment on the `prefix` that says this also affects `-alt` due to the slash being missing in the prefix.

pietroalbini (approved):

PR #66, opened 2021-04-17
Permit tagging with non-latest uploaded docker images

Mark-Simulacrum:


pietroalbini:
r=me once the input is declared in [`action.yml`](https://github.com/rust-lang/simpleinfra/blob/master/github-actions/upload-docker-image/action.yml]

PR #68, opened 2021-04-28
Update monitoring-related software on our instance

pietroalbini:
As we discussed during a past infra team meeting, this PR upgrades the Prometheus version so that in the future we can start using the Remote Write API. While I was at it I also upgraded other monitoring-related software versions:

* **Prometheus:** 2.12.0 => 2.26.0
  * No changes affecting us, nor deprecations
* **Grafana:** 6.3.3 => 7.5.5
  * [We'll probably need to tweak how we authenticate with AWS CloudWatch in production](https://grafana.com/docs/grafana/latest/installation/upgrading/#aws-cloudwatch-data-source). CloudWatch is only used for a few dashboards, so I can look at it after we upgrade.
  * There is a new UI theme (screenshots below)
* **Alertmanager:** 0.18.0 => 0.21.0
  * No breaking changes affecting us
* **node_exporter:** 1.0.1 => 1.1.2
  * There is a deprecation but it doesn't affect us

r? @Mark-Simulacrum 

---

| Grafana 6.3.3 | Grafana 7.5.5 |
| --- | --- |
| Screenshot 2021-04-28 at 16-55-46 Instance metrics - Grafana | Screenshot 2021-04-28 at 16-55-25 Instance metrics - Grafana |

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -6,14 +6,14 @@ global_collect_metrics_from:
 # Version of https://github.com/prometheus/node_exporter to install on our
 # instances. The checksum is the hash of the downloadable tarball for the
 # specified architecture, found in the GitHub releases.
-node_exporter_version: 1.0.1
+node_exporter_version: 1.1.2
 node_exporter_archs:
   x86_64:
     name: amd64
-    checksum: sha256:3369b76cd2b0ba678b6d618deab320e565c3d93ccb5c2a0d5db51a53857768ae
+    checksum: sha256:8c1f6a317457a658e0ae68ad710f6b4098db2cad10204649b51e3c043aa3e70d
   aarch64:
     name: arm64
-    checksum: sha256:017514906922fcc4b7d727655690787faed0562bc7a17aa9f72b0651cb1b47fb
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Everything looks reasonable to me, and this infra is not unusually interesting. It may be good to have a backup of the monitoring instance just so we can revert without losing our data though (some of the Prometheus changes are one-directional)

pietroalbini:
We should have regular backups for Grafana, I'll try a restore to make sure they work.

For Prometheus we never configured backups as the metrics are not that critical, I'll just stop the instance and do a copy of its storage before running the upgrade.

Mark-Simulacrum:
hm, I thought we would just do a snapshot of the system disk or so, but that works too

pietroalbini:
I would prefer to do a manual copy when Prometheus is stopped, I'm not sure how well it handles the disk being copied while it's running and honestly I don't think it's worth investigating that :)

Mark-Simulacrum:
Oh I meant, like, stopping the instance as a whole. Anyway I'll leave you to it - feel free to merge and deploy.

PR #70, opened 2021-05-27
Overwrite and unset passphrase variable

pierwill:
When the `with-rust-key` scripts exit, cleanup should include the data in the `passphrase` variable. We can do this by writing over the data then calling `unset`.


pietroalbini:
Discussed this on Zulip DMs and experimented a bit with Bash variables clearing, but this seems pretty fragile and prone to breakage. Thanks for all the investigation though!

PR #72, opened 2021-07-22
Redirect /rustup to the rustup book

jyn514:
Rustup lives on a different website than everything else. Make it easy to find
by adding a redirect to the main doc.rust-lang.org site.

See also https://github.com/rust-lang/rustup/issues/2714#issuecomment-818245894,
which proposes including the docs in releases themselves. This would have to
reverted if so.


jyn514:
The formatting changes look unrelated? I didn't modify the code they're pointing to, maybe a new version of teraform came out or something?

PR #75, opened 2021-09-20
Add checksum verification for `deploy.rs`

pierwill:
Since 1) we're downloading `deploy.rs` over the network, 2) it appears to handle secrets, and 3) the file doesn't seem to change very often, it makes sense to add an integrity check.


Mark-Simulacrum:
I don't think either the Travis config or Azure config is currently in use by any Rust repos, so we can probably just delete these. That said, I don't think hashes like this are a good idea - they turn updates to the script in an adventure in finding copies, which seems quite unfortunate.

PR #79, opened 2021-10-14
Add ansible config for dev desktops

oli-obk:
Discussion in https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/cloud.20dev.20desktops


ansible/roles/dev-desktop/templates/team_login/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+            "failed to create user"
+        );
+
+        // Get them ssh access
+        ensure!(
+            cmd("usermod", &["-a", "-G", "allow-ssh", &person.github])?.status.success(),
+            "failed to give user ssh access"
+        );
+    }
+    bail!("user not found");
```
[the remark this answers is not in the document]
oli-obk: true, but it should be very hard to hit the milliseconds during which your keyfile is invalid. I don't see how we can avoid that as sshd_config expects the file in a very specific location and I don't think we can specify a pattern.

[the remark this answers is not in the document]
Mark-Simulacrum: OK. We can see if there's pain in the future -- my worry would also be that if this is interrupted, a partial file maybe a security problem.

Maybe the way to go is the rename dance, but that seems likely insufficient without more investment so probably not worth it.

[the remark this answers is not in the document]
oli-obk: > I'll collect the list of users and compare it with the key files in the key directory and remove any that weren't on the list. That should not cause intermittent missing keys.

I implemented this

Mark-Simulacrum:
Pushed up a few commits with fixes/improvements while experimenting with the sample machine. One likely blocker I noted is removing the global Rust installation, which conflicts with usage of rustup (or at least rustup complains about it).

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -1,13 +1,15 @@
...
+- name: install Rust deps
   apt:
     name:
-      - cargo
       - pkg-config
       - libssl-dev
     state: present
 
+- name: install rustup in userspace for root
+  shell: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
Mark-Simulacrum, line 11: Can we pre-install rustup for each user by just copying these files or so into /etc/skel, perhaps?

I think it's something we have to do right away, otherwise the files don't get copied after the user is first created.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: install Rust deps
+  apt:
+    name:
+      - pkg-config
+      - libssl-dev
+    state: present
+
+- name: install rustup in userspace for root
+  shell: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
Mark-Simulacrum, line 11:
```
[WARNING]: Consider using the get_url or uri module rather than running 'curl'.  If you need to use command because get_url or uri is insufficient you can add 'warn: false' to this command task or set
'command_warnings=False' in ansible.cfg to get rid of this message.

fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"changed": true, "cmd": "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh", "delta": "0:00:00.359799", "end": "2021-12-15 18:30:07.489197", "msg": "non-zero return code", "rc": 1, "start": "2021-12-15 18:30:07.129398", "stderr": "info: downloading installer\nrustup: Unable to run interactively. Run with -y to accept defaults, --help for additional options", "stderr_lines": ["info: downloading installer", "rustup: Unable to run interactively. Run with -y to accept defaults, --help for additional options"], "stdout": "", "stdout_lines": []}
```

So probably a little more needs to be done to make this work.

oli-obk: I patched the command, it works locally, but not sure about ansible.

ansible/roles/dev-desktop/tasks/main.yml
oli-obk: Sorry, going on vacation now, I won't be able to do that. We can always run scripts for all users later if we want to do sth like that.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: install Rust deps
+  apt:
+    name:
+      - pkg-config
+      - libssl-dev
+    state: present
+
+- name: install rustup in userspace for root
+  shell: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | bash -s -- -y
```
pierwill, line 11: Let's avoid curling-to-bash. A simple hash verification would makes sense here:

```
set -eu

RUSTUP_VERSION=1.24.3
RUSTUP_SHA="a3cb081f88a6789d104518b30d4aa410009cd08c3822a1226991d6cf0442a0f8"

curl --proto '=https' --tlsv1.2 -sSf -O \
  https://raw.githubusercontent.com/rust-lang/rustup/${RUSTUP_VERSION}/rustup-init.sh

echo "${OUR_RUSTUP_SHA}  rustup-init.sh" | sha256sum --check --

sh rustup-init.sh \
   --default-toolchain nightly -y \
   --component rust-src
```

pierwill: I'm happy to submit a follow-up PR for this.

pierwill, line 11:
```suggestion
  shell: |
    set -eu
    RUSTUP_VERSION=1.24.3
    RUSTUP_SHA="a3cb081f88a6789d104518b30d4aa410009cd08c3822a1226991d6cf0442a0f8"
    curl --proto '=https' --tlsv1.2 -sSf -O \
      https://raw.githubusercontent.com/rust-lang/rustup/${RUSTUP_VERSION}/rustup-init.sh
    echo "${RUSTUP_SHA}  rustup-init.sh" | sha256sum --check --
    sh rustup-init.sh --default-toolchain nightly -y --component rust-src
```

oli-obk: Makes sense, thanks!

Mark-Simulacrum:
```
TASK [dev-desktop : Build team login cron job] **********************************************************************
fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"changed": true, "cmd": "cd /root/team_login && cargo build", "delta": "0:00:00.002152", "end": "2022-03-16 15:04:41.149915", "msg": "non-zero return code", "rc": 127, "start": "2022-03-16 15:04:41.147763", "stderr": "/bin/sh: 1: cargo: not found", "stderr_lines": ["/bin/sh: 1: cargo: not found"], "stdout": "", "stdout_lines": []}
```

It looks like we're not finding the rustup install -- maybe an explicit PATH is needed or something, not sure.

Mark-Simulacrum:
That did not seem to help; I'm still seeing the same error after pulling the latest changes.

Mark-Simulacrum:
```
TASK [dev-desktop : Build team login cron job] ***************************************************************************************************************************************************************
fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"changed": true, "cmd": "cd /root/team_login && PATH=$PATH:$HOME/.cargo/bin cargo build", "delta": "0:00:26.167428", "end": "2022-03-16 17:03:56.949407", "msg": "non-zero return code", "rc": 101, "start": "2022-03-16 17:03:30.781979", "stderr": "    Updating crates.io index\n Downloading crates ...\n  Downloaded openssl-sys v0.9.67\n  Downloaded autocfg v1.0.1\n  Downloaded unicode-xid v0.2.2\n  Downloaded proc-macro2 v1.0.29\n  Downloaded pkg-config v0.3.20\n  Downloaded socket2 v0.4.2\n  Downloaded ryu v1.0.5\n  Downloaded openssl-probe v0.1.4\n  Downloaded itoa v0.4.8\n  Downloaded quote v1.0.9\n  Downloaded syn v1.0.77\n  Downloaded curl v0.4.41\n  Downloaded libc v0.2.103\n  Downloaded cc v1.0.70\n  Downloaded libz-sys v1.1.3\n  Downloaded curl-sys v0.4.51+curl-7.80.0\n  Downloaded mini-internal v0.1.16\n  Downloaded miniserde v0.1.16\n   Compiling pkg-config v0.3.20\n   Compiling cc v1.0.70\n   Compiling libc v0.2.103\n   Compiling autocfg v1.0.1\n   Compiling proc-macro2 v1.0.29\n   Compiling unicode-xid v0.2.2\n   Compiling syn v1.0.77\n   Compiling ryu v1.0.5\n   Compiling curl v0.4.41\n   Compiling itoa v0.4.8\n   Compiling openssl-probe v0.1.4\nerror: linker `cc` not found\n  |\n  = note: No such file or directory (os error 2)\n\nerror: could not compile `curl` due to previous error\nwarning: build failed, waiting for other jobs to finish...\nerror: build failed", "stderr_lines": ["    Updating crates.io index", " Downloading crates ...", "  Downloaded openssl-sys v0.9.67", "  Downloaded autocfg v1.0.1", "  Downloaded unicode-xid v0.2.2", "  Downloaded proc-macro2 v1.0.29", "  Downloaded pkg-config v0.3.20", "  Downloaded socket2 v0.4.2", "  Downloaded ryu v1.0.5", "  Downloaded openssl-probe v0.1.4", "  Downloaded itoa v0.4.8", "  Downloaded quote v1.0.9", "  Downloaded syn v1.0.77", "  Downloaded curl v0.4.41", "  Downloaded libc v0.2.103", "  Downloaded cc v1.0.70", "  Downloaded libz-sys v1.1.3", "  Downloaded curl-sys v0.4.51+curl-7.80.0", "  Downloaded mini-internal v0.1.16", "  Downloaded miniserde v0.1.16", "   Compiling pkg-config v0.3.20", "   Compiling cc v1.0.70", "   Compiling libc v0.2.103", "   Compiling autocfg v1.0.1", "   Compiling proc-macro2 v1.0.29", "   Compiling unicode-xid v0.2.2", "   Compiling syn v1.0.77", "   Compiling ryu v1.0.5", "   Compiling curl v0.4.41", "   Compiling itoa v0.4.8", "   Compiling openssl-probe v0.1.4", "error: linker `cc` not found", "  |", "  = note: No such file or directory (os error 2)", "", "error: could not compile `curl` due to previous error", "warning: build failed, waiting for other jobs to finish...", "error: build failed"], "stdout": "", "stdout_lines": []}
```

Looks like this needs curl installed externally.

Mark-Simulacrum:
Couple notes:

* It looks like the message of the day doesn't quite work, or at least logging in myself I'm not seeing it yet
* Install the team login binary when run through ansible after the initial deploy can fail with "cannot create regular file" if the binary is currently running. Ideally we can seamlessly redeploy as needed without impacting existing users, so this seems like it's good to fix. (Particularly if we e.g. modify the source code of the team login script).
* the "common : remove old group memberships" task in Ansible will drop allow-ssh from a bunch of users that get created by the cron job - I'm not sure what the right fix here is, but presumably this means that deploying ansible will erase a bunch of permissions, which seems not great. (They will get restored in ~5 minutes, but still.)

oli-obk:
> * It looks like the message of the day doesn't quite work, or at least logging in myself I'm not seeing it yet

Yea I was worried about that. The default sshd config disables motd and I'm not sure if reenabling it actually works in this setup. We could change the default sshd config, but that would change it for all servers.

 

> * Install the team login binary when run through ansible after the initial deploy can fail with "cannot create regular file" if the binary is currently running.

Huh? I thought linux can overwrite binaries that are running and that there are no locks on them? I'll look into that. Maybe we just spawn a script that keeps trying to copy the new binary until that succeeds.



> * the "common : remove old group memberships" task in Ansible will drop allow-ssh from a bunch of users that get created by the cron job -

Launch the cron script once manually right after deploy?

Mark-Simulacrum:
> Yea I was worried about that. The default sshd config disables motd and I'm not sure if reenabling it actually works in this setup. We could change the default sshd config, but that would change it for all servers.

I do see *some* message of the day, I think, but seems like something to fix. I don't think the particular fix is too important -- if we need to enable messages of the day across all servers, that seems OK, for example.

> Huh? I thought linux can overwrite binaries that are running and that there are no locks on them? I'll look into that. Maybe we just spawn a script that keeps trying to copy the new binary until that succeeds.

```
fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"changed": true, "cmd": "cp /root/team_login/target/debug/team_login /etc/cron.team_login", "delta": "0:00:00.002736", "end": "2022-03-16 19:55:11.805673", "msg": "non-zero return code", "rc": 1, "start": "2022-03-16 19:55:11.802937", "stderr": "cp: cannot create regular file '/etc/cron.team_login': Text file busy", "stderr_lines": ["cp: cannot create regular file '/etc/cron.team_login': Text file busy"], "stdout": "", "stdout_lines": []}
```

> Launch the cron script once manually right after deploy?

I think we probably want to not run the drop groups task for this server, and potentially that means it needs a different common base (or parts of the common base need factoring out); it's our first server with a non-empty set of additional users that aren't just the infra admins and aren't created by Ansible.

(An alternative approach could be to move the user creation out of the server cron job and into Ansible, and then add logic to e.g. rust-lang/sync-team to run Ansible to deploy changes across servers after team repo updates. But that is a larger change, for sure.)

Mark-Simulacrum:
```
TASK [dev-desktop : Set up the team login cron job] 
fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"msg": "The conditional check 'task_result.rc == 0' failed. The error was: error while evaluating conditional (task_result.rc == 0): 'dict object' has no attribute 'rc'"}
```

Mark-Simulacrum:
OK, that seemed to work in terms of deploying. I'm not yet seeing the message of the day this is trying to configure though:

```
Welcome to Ubuntu 20.04.4 LTS (GNU/Linux 5.13.0-1017-aws x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Thu Mar 17 16:51:48 UTC 2022

  System load:  0.05               Processes:             364
  Usage of /:   2.2% of 193.82GB   Users logged in:       0
  Memory usage: 1%                 IPv4 address for ens5: 10.0.0.85
  Swap usage:   0%

 * Ubuntu Pro delivers the most comprehensive open source security and
   compliance features.

   https://ubuntu.com/aws/pro

21 updates can be applied immediately.
3 of these updates are standard security updates.
To see these additional updates run: apt list --upgradable


*** System restart required ***
Last login: Wed Mar 16 19:54:09 2022 from ...
```

oli-obk:
could you remove the IP address lockdown? That would make it easier and we'll soon have to do it anyway

Mark-Simulacrum:
I don't think it's currently in place already.

Mark-Simulacrum:
OK, that seems to have worked:

```
Continued use of this server implies that you accept the rules.
You can find the rules at
https://foundation.rust-lang.org/policies/cloud-compute-program/
Last login: Thu Mar 17 17:24:33 2022 from ...
simulacrum@dev-desktop:~$
```

I think this is nearly ready to go then, probably we just need to button up the policy and confirm with Joel that this will work as the only/primary prompt users get.

@pietroalbini might also want to take a look at the Terraform and Ansible configurations, but I figure in the alpha/beta period with just a few testers the current machine should likely be OK.

pietroalbini:
I'll take a look in the coming days.

Mark-Simulacrum:
We are getting an alert from our monitoring infra:

> motd-news.service on dev-desktop failed

Can we disable that systemd service or otherwise make it not actually fail to run?

ansible/roles/common/tasks/users.yml
```diff
@@ -58,6 +58,7 @@
 # remove them.
 - name: remove old group memberships
   command: "gpasswd -d {{ item.member | quote }} {{ item.group | quote }}"
+  when: not keep_old_user_permissions_on_redeploy
```
pietroalbini, line 61: I don't think we should just disable this step, as that would prevent removing sudo users (that are only managed in ansible) and practically result in us forgetting to remove those perms.

A better approach here would be to add (only on the dev-desktop machines) [an extra group of users allowed to ssh](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/common/templates/ssh/sshd_config#L37) called something like `dev-desktop-allow-ssh`. That way it won't interfere with the ansible group management.

ansible/roles/dev-desktop/files/team_login/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Check if user exists
+        let id = cmd("id", &[&person.github]).expect("failed to run `id` command");
+        if id.status.success() {
+            continue;
+        }
+
+        // If user does not exist, create it
+        assert!(
+            cmd("useradd", &["--create-home", &person.github])?
```
pietroalbini, line 72: Can we have a prefix like `gh-{username}` to prevent users from colliding with system users?

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+  # if the cron job is running right now, keep retrying until it finishes
+  register: task_result
+  until: task_result is not failed
+  retries: 10
+  delay: 5
+
+- name: Set up the files that initially appear in a user's home dir
+  copy:
+    src: skel/
+    dest: /etc/skel/
```
pietroalbini, line 55: I'm not too keen on the `skel` approach for providing pre-made scripts, as making any update/fix to them is going to be a pain for existing users on a machine. I'd directly add them to `/usr/local/bin`, that way they're centrally managed and we can add/update them without any issue.

terraform/dev-desktops/instance.tf
```diff
@@ -0,0 +1,143 @@
...
+  type    = "A"
+  records = [aws_eip.dev_desktop.public_ip]
+  ttl     = 60
+}
+
+// Create the IAM role used by the dev-desktop.
+
+// Create the EC2 instance itself.
+
+data "aws_ami" "ubuntu_bionic" {
```
pietroalbini, line 101: Nit: 20.04 is focal, not bionic :slightly_smiling_face: Removing the codename altogether would probably be better.

ansible/roles/dev-desktop/files/team_login/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    let data = String::from_utf8(data).unwrap();
+    let all: All = miniserde::json::from_str(&data).unwrap();
+    let mut users = HashSet::new();
+    for person in all.members {
+        users.insert(person.github.clone());
+        // Get the keys the user added to their github account.
+        // Do this every time the cronjob runs so that they get their new keys if necessary.
+        let mut keys = Vec::new();
+        handle
+            .url(&format!("https://github.com/{}.keys", person.github))
```
pietroalbini, line 48: This is going to result in a lot of requests every 5 minutes, do we know what the rate limit on the endpoint is?

ansible/roles/dev-desktop/files/team_login/src/main.rs
```diff
@@ -0,0 +1,105 @@
+const TEAM_URL: &str = "https://team-api.infra.rust-lang.org/v1/teams/all.json";
```
pietroalbini, line 1: I'd add a new `dev-desktop = true` permission rather than just allowing the `all` team. Mostly because that'd allow to manually set `dev-desktop = false` on any user or team if problems arise.

pietroalbini (commented):
Left some comments.

In general, I'd still prefer for the sync script not to be in Rust but to be in either bash or even better python: it'd remove the need to manage the rustup installation on the root user, the problem with replacing binaries, and it'd make on the fly changes easier.

If you don't have the time to do it I can push a commit changing the script over to Python.

ansible/roles/dev-desktop/files/team_login/src/main.rs
pietroalbini: An example of colliding user here would be `aidanhs`, but I could also imagine some avenues for privilege escalation.

pietroalbini:
Also, how do we expect users to authenticate with GitHub to push code? I might've missed that discussion.

oli-obk:
> Also, how do we expect users to authenticate with GitHub to push code? I might've missed that discussion.

they will have to use ssh agent forwarding for now

oli-obk:
> the problem with replacing binaries

that turned out to be a problem replacing a cron.d configuration file, not a binary, I misdiagnosed

terraform/dev-desktops/instance.tf
pietroalbini: No, this was mostly for Mark (who authored the code). I can probably also make those changes eventually. It's not a bit deal at all.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Set up message of the day
+  template:
+    src: motd_rules
+    dest: /etc/motd
+
+- name: Install the team login binary
+  shell: "cp {{ vars_team_login_path }}/target/debug/team_login /etc/cron.team_login"
+
+- name: Set up the team login cron job
```
pietroalbini, line 45: Not a blocker for the initial prototyping, but you mentioning the sync is currently failing due to the team repo changes not being applied reminded me that we don't have proper monitoring in place for cronjobs failing, but we do have monitoring in place for systemd units failing.

Could the cronjob be switched to a systemd timer/service combo? You can see an example in this repo, with the corresponding [.service](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/docker/templates/update-images/docker-images-update.service) and [.timer](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/docker/templates/update-images/docker-images-update.timer).

ansible/roles/dev-desktop/tasks/main.yml
pietroalbini: Note that you also moved the `config.toml` to `/usr/local/bin`, even though it's not a script :sweat_smile:

I'm not sure exactly what to do for that, either omit it or still add it as part of the skeleton.

pietroalbini:
Thanks for the fixes! I left a comment in https://github.com/rust-lang/team/pull/720 with an alternative approach that integrates better with the rest of the team repo.

> they will have to use ssh agent forwarding for now

Hmm, that makes me worried security wise. Encouraging the use of SSH forwarding would mean that root on those machines would be able to use the SSH keys of all connected team members, which could lead to security incidents.

I'm thinking both of a vulnerability allowing privilege escalation on the system, but also just the liability of infra-admins *theoretically* being able to do such thing. If someone's GitHub account gets compromised I'd prefer to say "I didn't have access to those keys" rather than "I didn't abuse the access I have".

I know it's much more of a pain, but I'd prefer if team members did something like this on their local machines to push code to GitHub (pulling code from the dev machine and then pushing it to GitHub):

```
git fetch gh-pietroalbini@dev-desktop.infra.rust-lang.org:rust.git mybranch
git push origin mybranch
```

If we go that route we will also need to set `AllowAgentForwarding no` in `sshd_config` to nudge users they should really not do that (even though it's bypassable if someone is stubborn enough).

What do you think?

JoelMarcey:
> they will have to use ssh agent forwarding for now

This got me thinking -- is there (or should we have) some sort of brief documentation on how to use the provided cloud services? I could see such documentation having a statement about authenticating with GitHub, maybe through a FAQ or something.

Or will initial usage just be obvious?

oli-obk:
> I know it's much more of a pain

I used to have this. It's not great at all. We're not going to nudge users to using that, we're going to nudge them to add private SSH keys to the cloud desktops.

Especially considering that many people will use vscode remote to work on these machines, so they would need a separate terminal just for pushing their state. And then they need to remember which checkout they want to push... instead of just working in the directory they are in.

Covenience wins over security for most people.

I tried some approaches around github authentication, which can be limited to pushing to specific repos. It's promising, but I haven't made it convenient enough yet.

pietroalbini:
I still have deep concerns about the security and liability angle of forwarding the SSH agents inside the machine.

With developers forwarding their SSH agents into the machine, anyone with sudo access will be able to use that agent to access any private personal repository, any *work* repository for all companies that are not on GitHub Enterprise Cloud with SAML, to [request disabling GitHub 2FA](https://docs.github.com/en/authentication/securing-your-account-with-two-factor-authentication-2fa/recovering-your-account-if-you-lose-your-2fa-credentials#authenticating-with-a-verified-device-ssh-token-or-personal-access-token), and to perform any activity outside GitHub authenticating with those keys.

It is a huge liability both for the foundation and for the individual volunteers on the infrastructure team with sudo access on the machine. I have full trust nobody on the infra team will abuse sudo access, but any privilege escalation vulnerability or compromise of an infra team member's credentials will allow an attacker to gain immense amounts access.

If we go ahead with this, on the security side sudo access on the machine will have to be extremely locked down, with at most a couple people (of infra-admins) with sudo access on the machine, and we're going to need to be *extremely* diligent at updating those machines, as any privilege escalation could be catastrophic. I'm not sure we have the human bandwidth to be that diligent on updates as of today to be honest, and I'd still not be comfortable with this even if we had that.

---

I have some thoughts on how to handle authentication to GitHub that should hopefully have no friction and should be relatively easy to implement.

[GitHub Apps](https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps) can be installed on individual accounts or even individual repositories, thus scoping down the amount of access they have and limiting the blast radius of a server compromise only to the repossitories developers explicitly allowed to be accessed by the dev machines. We can tell users to first install the GitHub App granting it access to the repositories it should be able to access.

On the server, we can create a [git credential provider](https://git-scm.com/docs/gitcredentials) that requests an ["installation token"](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation) from GitHub acting as the user's installation of the app, and [use that token for all git operations](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#http-based-git-access-by-an-installation). The git credential provider is needed as those tokens are short-lived, so even if we cache them we're going to need to regenerate them often.

oli-obk:
> On the server, we can create a [git credential provider](https://git-scm.com/docs/gitcredentials) that requests an ["installation token"](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation) from GitHub acting as the user's installation of the app, and [use that token for all git operations](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#http-based-git-access-by-an-installation). The git credential provider is needed as those tokens are short-lived, so even if we cache them we're going to need to regenerate them often.

Not sure how this changes things. Someone who has root can still just do their operations while a token is valid. Sure it's not the entire time someone is logged in, but instead just the short time while people have to do some operations. They also have to repeat to do some login operation every time they connect.

As I said before, I am *very* confident any friction like this will just lead to people uploading some private ssh keys or adding permanent tokens for the repos that they want to access from the cloud desktops. This is obviously the worst outcome, and I very much want to avoid it.

I am aware of the security problem and have been trying to figure it out before the cloud desktop proposal was even written.

We can obfuscate and stuff, but we can't possibly make it truly secure even by using single-use tokens (though in the latter case we at least notice when someone is using them).

The only true way, as you suggested earlier, is to always sync to local and do the push from there. And maybe that is the right way, but before we fall back to this we need to figure out a way to at least make it easy. Ideally we'd have a local program running that is accessible via a forwarded port from the cloud desktop. Then we hook into the git operations and insert the credentials transparently.

That is a lot of work (I have not found any existing tools for this), so I am fine with asking for funding for someone to develop this solution or some other solution that I can't see right now. But *right now* I am strongly in favour of just going with it.

oli-obk:
I completely misunderstood how the github App would work. pietro explained it to me on a private chat and I am confident this will work without causing any friction for users. The only thing we need to figure out is how to report good errors when the user hasn't set up a repo in the app yet.

I'll look into creating a github App for the cloud desktops now.

oli-obk:
Ok... turns out if you use vscode remote, you can just use the vscode github app, which works even better than what we want afaict, as it runs on the user side and gives out tokens to the server without ever having long term secrets on the server

oli-obk:
I guess we can call it a day and not support schemes other than vscode remote in the first version?

Mark-Simulacrum:
I presume giving out tokens to the server is basically equivalent threat model wise to Pietro's concern around forwarded ssh agents? Unless those tokens are scoped to a repository, which would surprise me since typically you can't do that IIRC?

It seems to me that we should still pursue the App model if that seems like it addresses this concern. (Or some other similar alternative).

In general I suspect we'll still get folks using things that technically expose them to sudo users using their credentials; I'm not sure we can entirely prevent that under the current shared machine model. I think it's probably OK for the alpha mode where we try this out to gain user experience reports, but we may wish to consider moving to alternative models as we go into beta etc (for example, one could imagine using some kind of cloud-level shared tenancy, though I would need to research to know what's possible).

oli-obk:
> It seems to me that we should still pursue the App model if that seems like it addresses this concern.

vscode uses the app model. When you register the vscode app, you choose which repos it may push to.

If we pursue our own app thing, it's actually worse. root can just push to any repo of any user that activated the app, irrespective of whether they are logged in or not.

Mark-Simulacrum:
Ah, interesting. Well, then that solution likely should be recommended for vscode users, but I suspect that's a subset (though maybe a large one) of the users interested in this.

I'm not sure I entirely follow why vscode is different from our own equivalent model - presumably we could also have a wrapper client (equivalent to vscode) around user's ssh invocation which provisions the token temporarily or so?

oli-obk:
> presumably we could also have a wrapper client (equivalent to vscode) around user's ssh invocation which provisions the token temporarily or so?

yea, we can definitely do that, but I'm already in very murky waters here with me eyeballing this to be secure and writing random scripts that should hopefully work until someone breaks the python dependencies we use or github does random things to its API.

Writing such scripts to work on every operating system wrapping a ssh connection seems... not great to maintain either.

JoelMarcey:
I think in order to limit the variables that may exist for the first test run of access to the machines, it is not unreasonable to ask our initial set of beta testers to try accessing with VS Code just so we can verify that the underlying system works and doesn't break. And then when we determine we are on good footing there, we can figure out how to expand usage of the service to support more clients.

pietroalbini:
I'll take another look at this once I'm done with the 1.60.0 work.

ansible/roles/dev-desktop/templates/motd_rules
```diff
@@ -0,0 +1,3 @@
+Continued use of this server implies that you accept the rules.
+You can find the rules at
+https://foundation.rust-lang.org/policies/cloud-compute-program/
```
JoelMarcey, line 3:
```suggestion
By accessing and making use of this Cloud Compute Program resource, you are agreeing that you have both read and will abide by its official Access Agreement. The Access Agreement can be found at https://foundation.rust-lang.org/policies/cloud-compute-program/

```

pietroalbini:
> If we pursue our own app thing, it's actually worse. root can just push to any repo of any user that activated the app, irrespective of whether they are logged in or not.

I don't think that distinction matters in practice. If an attacker gains root access they have countless ways to detect when the user logs in and immediately steal their token, without being discovered beforehand.

Liability and security wise I'm comfortable with our own GitHub App model for non-vscode users.

We can also prevent too much damage if users select "install for all my repositories": when the users try to obtain a token, we can [list all the repositories they granted access to](https://docs.github.com/en/rest/reference/apps#list-repositories-accessible-to-the-app-installation) and abort if any private repository is allowed, telling them to manually select which repositories to grant access to. There are also other heuristics we could implement (repositories with more than N stars or non-forks), but that could lead to false positives and we'd need an "I know what I'm doing" opt out.

> yea, we can definitely do that, but I'm already in very murky waters here with me eyeballing this to be secure and writing random scripts that should hopefully work until someone breaks the python dependencies we use or github does random things to its API.
>
> Writing such scripts to work on every operating system wrapping a ssh connection seems... not great to maintain either.

I agree we shouldn't develop scripts to run on the client side.

JoelMarcey:
So I think all that is left here is:

1. First prototype of [GitHub App authentication](https://github.com/rust-lang/simpleinfra/pull/79#issuecomment-1087396763)
2. Disable the [systemd service](https://github.com/rust-lang/simpleinfra/pull/79#issuecomment-1072496598), if still required.
3. Rebuild and redeploy of the instance
4. Ship the Alpha!

oli-obk:
I'm going on vacation starting Friday. I won't be able to finish  the custom auth app tomorrow. Let's keep do the alpha with vscode only and add the custom auth app later

JoelMarcey:
@oli-obk If we just use vscode only for now, do we then just have to:

1. Disable the [system service](https://github.com/rust-lang/simpleinfra/pull/79#issuecomment-1072496598), if still required.
2. Rebuild and redeploy of the instance
3. Ship the Alpha!

And then we can add the GitHub Authentication mid-alpha when you return? (Btw, I assume you are best suited to add that authentication and it would require a bit of learning curve for someone else to take that on?)

oli-obk:
The github part is done. I'll have to figure out the git auth integration from scratch, too, so there's no benefit in me doing it. I'll pick it up on the 26th if no one grabbed it by then

JoelMarcey:
Have a good vacation @oli-obk !

ansible/roles/dev-desktop/files/team_login/src/main.rs
```diff
@@ -36,16 +38,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let all: All = miniserde::json::from_str(&data).unwrap();
     let mut users = HashSet::new();
     for username in all.github_users {
+        let url = format!("https://github.com/{}.keys", username);
```
oli-obk, line 41: facepalm emoji

ansible/roles/dev-desktop/files/skel/config.toml
```diff
@@ -0,0 +1,9 @@
+changelog-seen = 2
+
+# Change this to the profile of your choice
+profile = "compiler"
+
+[rust]
+# Build with debug assertions so we get useful RUSTC_LOG
+# and notice debug assertions failing.
+debug = "true"
```
JoelMarcey, line 9:
```suggestion
debug = true
```

so we don't get.....

```
failed to parse TOML configuration 'config.toml': invalid type: string "true", expected a boolean for key `rust.debug`
```

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+      - build-essential
+      - libssl-dev
+      - python3-pip
+
+    state: present
+
+- name: install python libs
+  shell: "pip3 install PyGithub requests cryptography"
+
+- name: install rustup in userspace for root
```
JoelMarcey, line 22: in the current instance build, `rustup` doesn't look to be working, even though it appears to be installed here.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+
+- name: Configure update script service
+  copy:
+    src: team_login/
+    dest: "{{ vars_team_login_path }}"
+
+- name: Build team login cron job
+  shell: "cd {{ vars_team_login_path }} && PATH=$PATH:$HOME/.cargo/bin cargo build"
+
+- name: Enable message of the day
```
JoelMarcey, line 40: This is not currently showing on the current instance build based on this config.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
+---
```
JoelMarcey, line 1: Can we make the default shell `bash` instead of `sh`?

JoelMarcey (commented):
A couple of things after @yaahc and I gave this a whirl.

oli-obk:
The scripts for handling rust workspaces are currently only available for root, we need to make them executable by everyone

Mark-Simulacrum:
It looks like the motd work is breaking the fwupd-refresh.service on the machine:

```
● fwupd-refresh.service - Refresh fwupd metadata and update motd
     Loaded: loaded (/lib/systemd/system/fwupd-refresh.service; static; vendor preset: disabled)
     Active: failed (Result: exit-code) since Sun 2022-04-24 12:48:59 UTC; 2h 38min ago
TriggeredBy: ● fwupd-refresh.timer
       Docs: man:fwupdmgr(1)
    Process: 499412 ExecStart=/usr/bin/fwupdmgr refresh (code=exited, status=1/FAILURE)
   Main PID: 499412 (code=exited, status=1/FAILURE)

Apr 24 12:48:59 dev-desktop.infra.rust-lang.org systemd[1]: Starting Refresh fwupd metadata and update motd...
Apr 24 12:48:59 dev-desktop.infra.rust-lang.org systemd[1]: fwupd-refresh.service: Main process exited, code=exited, status=1/FAILURE
Apr 24 12:48:59 dev-desktop.infra.rust-lang.org systemd[1]: fwupd-refresh.service: Failed with result 'exit-code'.
Apr 24 12:48:59 dev-desktop.infra.rust-lang.org systemd[1]: Failed to start Refresh fwupd metadata and update motd.
```

I'm not sure why, but maybe something to look into. I've temporarily stopped that service and timer manually:

```
sudo systemctl reset-failed fwupd-refresh.service
sudo systemctl disable fwupd-refresh.timer
sudo systemctl stop fwupd-refresh.timer
```

ansible/roles/dev-desktop/tasks/main.yml
oli-obk: rustup needs to be installed per user. There is an open issue to support multi-user environments: https://github.com/rust-lang/rustup/issues/313

ansible/roles/dev-desktop/tasks/main.yml
oli-obk: My default shell after login is

oli-obk: Ah interesting, only the admin users have this, everyone else has `sh` 😕 looking into it

JoelMarcey:
Love the explainer! 👍

Can we land this now? Or, at least, rebuild the instance to include all these new improvements?

Mark-Simulacrum:
I'll rebuild the instance later tonight or tomorrow morning from scratch.

Mark-Simulacrum:
```
TASK [dev-desktop : Disable automatically generated message of the day] ***************************************************************************
[WARNING]: Consider using the file module with state=absent rather than running 'rm'.  If you need to use command because file is insufficient you
can add 'warn: false' to this command task or set 'command_warnings=False' in ansible.cfg to get rid of this message.
fatal: [dev-desktop.infra.rust-lang.org]: FAILED! => {"changed": true, "cmd": "rm -r /etc/update-motd.d/*", "delta": "0:00:00.002726", "end": "2022-04-26 00:40:52.538122", "msg": "non-zero return code", "rc": 1, "start": "2022-04-26 00:40:52.535396", "stderr": "rm: cannot remove '/etc/update-motd.d/*': No such file or directory", "stderr_lines": ["rm: cannot remove '/etc/update-motd.d/*': No such file or directory"], "stdout": "", "stdout_lines": []}
```

Mark-Simulacrum:
Fixing that with a `-f` flag to rm (probably not the right fix, but not sure, see also the warning) seems to have completed the rest of the tasks -- should be deployed.

ansible/roles/dev-desktop/tasks/main.yml
oli-obk: One problem: all existing users now are stuck with `sh` 😆 I'll patch them manually, but in the future all deploys will use bash

oli-obk: Done, you may need to delete the `.vscode-server` folder on the server as it remembers the shell at setup time out of some reason

Mark-Simulacrum: We'll definitely recreate the base machine (not just Ansible deploy) before the next stage, so that's a temporary condition.

Mark-Simulacrum:
Fired: motd-news.service on dev-desktop.infra.rust-lang.org:9100 failed is still logging into our alerts after the deploy.

Mark-Simulacrum:
Disabling the service is indeed not enough, you also need to disable and stop the timer:

```
● motd-news.timer - Message of the Day
     Loaded: loaded (/lib/systemd/system/motd-news.timer; enabled; vendor preset: enabled)
     Active: active (waiting) since Wed 2022-04-27 11:42:27 UTC; 3min 8s ago
    Trigger: Wed 2022-04-27 13:57:14 UTC; 2h 11min left
   Triggers: ● motd-news.service
```

(I did that manually after recreating the instance)

ansible/roles/common/tasks/users.yml
pietroalbini: @oli-obk this is still pending btw :slightly_smiling_face:

ansible/roles/dev-desktop/files/github_app/readme.md
```diff
@@ -0,0 +1,28 @@
+# How to setup an App
```
JoelMarcey, line 1: Thanks for this @oli-obk ! 

Just wanted to clarify something -- will the user (i.e. those that are logging into the cloud compute machine to do Rust development) be setting up the App? Or will an admin of the cloud compute machine set up the App and then admin installs and grants access to the users?

In other words, does a user or admin follow these steps? I feel like it is an admin, but wasn't 100% sure.

oli-obk: Yea, this is an admin. I mean, we could make it an app per user, but that's annoying to set up and not really more secure unless you wipe your key after your session and copy it again when you reconnect.

The only step the user has to follow with this is to install the app, for which we'll provide them with a URL, not sure when yet, possibly at login when they haven't set up the app yet

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -0,0 +1,137 @@
...
+    dest: /etc/profile.d/dev-desktop-credential-path.sh
+    content: 'PATH=$PATH:/home/github-app-credentials'
+
+- name: Insert the git credential login for everyone
+  copy:
+    src: templates/gitconfig
+    dest: /etc/gitconfig
+
+# TODO:
+#- name: Set up the github app key- and id-file
```
oli-obk, line 137: help! We've reached the point where someone needs to create the app and secrets, store the secrets somewhere secret and set it up so ansible can find them and upload them. I can't do any of these steps and it's probably faster for one of you to set it up than to explain it to me (and me learning it has no benefit, because it's unlikely I'll be handling infra-secrets in the future).

oli-obk:
modulo bugs (likely, will need to see after redeploy), I think I have implemented and addressed everything I can.

ansible/roles/dev-desktop/files/github_app/git-credential-dev-desktop
```diff
@@ -0,0 +1,32 @@
...
+
+config = [x.strip() for x in stdin]
+
+if config[1] != "host=github.com":
+    exit(0)
+
+path = config[2].split('=')
+path = path[1].split('/')
+
+user = path[0]
```
oli-obk, line 17: oh this is the wrong user. we need to get the user from the logged in username, otherwise users can steal other user's repos. will fix this tomorrow

ansible/roles/common/templates/ssh/sshd_config
```diff
@@ -34,7 +34,7 @@ AcceptEnv LANG LC_*
 
 Subsystem sftp /usr/lib/openssh/sftp-server
 
-AllowGroups allow-ssh
+AllowGroups allow-ssh dev-desktop-allow-ssh
```
pietroalbini, line 37: Can we only add this on the dev-desktop machines rather than all machines? Maybe adding an `extra-ssh-allow-groups` variable.

ansible/roles/dev-desktop/files/github_app/readme.md
```diff
@@ -0,0 +1,53 @@
...
+We're using the python github library for all github operations.
+You can find the docs at https://pygithub.readthedocs.io/en/latest/introduction.html
+
+# How to setup an App
+
+1. Go to https://github.com/settings/apps
+2. New Github App
+3. Fill out metadata (name and url)
+4. disable WebHook checkbox
+5. Set `Contents - Repository contents, commits, branches, downloads, releases, and merges.` to read/write
```
pietroalbini, line 14: The workflows permission also needs to be granted, otherwise it's not possible to push GitHub Actions workflow files.

ansible/roles/dev-desktop/files/github_app/readme.md
```diff
@@ -0,0 +1,53 @@
+These instructions are for server-side setup and debugging.
```
pietroalbini, line 1: We usually have all our docs in the [Documentation section of the forge](https://forge.rust-lang.org/infra/docs/index.html), could this be moved to a PR there? I'd create separate pages for team members and for the infra team (like for AWS access).

oli-obk: This documentation is purely for the infra team, there's nothing there a user of the dev desktops needs to read

ansible/roles/common/templates/ssh/sshd_config
oli-obk: I don't know how to do that.

Appending sshd_config with files in sshd_config.d has been very weird with overwriting previous flags. I'll try it out, but if that doesn't work, we should imo just allow this group or rename it to something generic like `allow-ssh-noninfra`

ansible/roles/dev-desktop/files/github_app/readme.md
pietroalbini: Definitely! We're trying to keep the documentation on the infrastructure on the forge though, so that it's all in a central place rather than being scattered around :slightly_smiling_face:

ansible/roles/common/templates/ssh/sshd_config
pietroalbini: I would add this to the [default variables for the Ansible role](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/common/defaults/main.yml):

```yaml
allow_ssh_extra_groups: ""
```

Then the current line in the `sshd_config` can be replaced with:

```
AllowGroups allow-ssh {{ allow_ssh_extra_groups }}
```

And finally in the playbook, in the common role we add:

```yaml
- role: common
  # ...
  allow_ssh_extra_groups: dev-desktop-allow-ssh
```

I agree it's not worth it messing with `sshd_config.d`.

ansible/roles/dev-desktop/files/github_app/readme.md
oli-obk: Oh sorry, I'm all for that. I was just confused by what you wanted me to split when you wrote

> I'd create separate pages for team members and for the infra team

ansible/roles/common/templates/ssh/sshd_config
oli-obk: Oooh, these templates are great. Thanks for the explanation

Mark-Simulacrum:
```
● fwupd-refresh.service - Refresh fwupd metadata and update motd
     Loaded: loaded (/lib/systemd/system/fwupd-refresh.service; static; vendor preset: disabled)
     Active: failed (Result: exit-code) since Fri 2022-05-06 10:43:05 UTC; 11h ago
TriggeredBy: ● fwupd-refresh.timer
       Docs: man:fwupdmgr(1)
    Process: 1388285 ExecStart=/usr/bin/fwupdmgr refresh (code=exited, status=1/FAILURE)
   Main PID: 1388285 (code=exited, status=1/FAILURE)

May 06 10:43:05 dev-desktop.infra.rust-lang.org systemd[1]: Starting Refresh fwupd metadata and update motd...
May 06 10:43:05 dev-desktop.infra.rust-lang.org systemd[1]: fwupd-refresh.service: Main process exited, code=exited, status=1/FAILURE
May 06 10:43:05 dev-desktop.infra.rust-lang.org systemd[1]: fwupd-refresh.service: Failed with result 'exit-code'.
May 06 10:43:05 dev-desktop.infra.rust-lang.org systemd[1]: Failed to start Refresh fwupd metadata and update motd.
```

Probably as a result of our MOTD changes, I think pre-existing (maybe I've even noted it already).

pietroalbini (approved):

pietroalbini:
Merging this, we can iterate with other PRs.

PR #87, opened 2021-12-10
Add a few release process scripts

Mark-Simulacrum:
This adds the two more interesting blocks of code, with some slight alterations, from https://forge.rust-lang.org/release/process.html to this repository. My hope is that we can use this as a baseline step for the next release, and then hopefully gradually start moving this infrastructure into the remote release process (i.e., start-release.py's invoked cloud build resources). Even in this repository, I think it'll be much nicer to invoke scripts rather than copy/pasting line-by-line and hoping to not get anything wrong.

Obviously, there are more release steps that this doesn't yet cover, but it seems like a good start, and covers the ones I worry about the most -- at least to some extent.

r? @pietroalbini

### DIFF
--- a/release-scripts/master-to-beta.sh
+++ b/release-scripts/master-to-beta.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+if [ ! -e x.py ]; then
+    echo "Should be run from a rust-lang/rust checkout"
+    exit 1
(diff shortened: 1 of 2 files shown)


pietroalbini (approved):

Mark-Simulacrum:
I'll file a PR updating forge soon, as well.

PR #93, opened 2022-04-25
Tweaks from when theory met practice when deploying the playground

shepmaster:


ansible/envs/dev-example/group_vars/playground.yml
```diff
@@ -2,6 +2,8 @@
 
 vars_playground_domain: example.com
 
+vars_playground_s3_bucket: playground-artifacts-i32
```
pietroalbini, line 5:
```suggestion
vars_playground_s3_bucket: bucket-name
```

PR #94, opened 2022-05-21
Improve authentication with the GitHub App to dev-desktop

pietroalbini:
This PR improves `git-credential-dev-desktop` to the point that it successfully works (even though the error messages are still quite poor when GitHub permissions are missing):

* The scripts have been moved to `/usr/local/bin` and the confiuration to `/etc/github-app-credentials`.
* The two scripts have been merged together, removing the CLI to get the token of an arbitrary user.
* Permissions have been fixed.
* A type error in the script has been fixed.


ansible/roles/dev-desktop/files/git-credential-dev-desktop-inner
```diff
@@ -0,0 +1,146 @@
...
+
+    if "host" not in config or config["host"] != "github.com":
+        exit(0)
+    git_user, repo = config["path"].split("/", 1)
+    if repo.endswith(".git"):
+        repo = repo[:-len(".git")]
+
+    # Get the username of the caller, not the owner of
+    # this script (which is setuid)
+    user = os.getlogin()
```
bjorn3, line 80: From https://linux.die.net/man/3/getlogin_r (emphasis mine)

> Unfortunately, it is often rather easy to fool getlogin(). Sometimes it does not work at all, because some program messed up the utmp file. Often, it gives only the first 8 characters of the login name. The user currently logged in on the controlling terminal of our program need not be the user who started it. **Avoid getlogin() for security-related purposes.**

pietroalbini: Rebased on top of master, and added the last commit to address this. I am *pretty sure* that's safe, but I'd like another set of eyes on it.

jdno (commented):
I've tested this branch on a virtual machine and functionally everything works. 👍

As far as I can tell, the current implementation using `sudo` makes sense. I re-read the documentation on [sudo](https://linux.die.net/man/8/sudo) and [sudoers](https://linux.die.net/man/5/sudoers), and the script matches the expected behavior outlined in the docs. As did playing around with `sudo` to confirm that the `SUDO_USER` variable is set correctly. But I'm not a security expert, so not sure if we want to ask someone else to review as well...

jdno (approved):
👍

PR #103, opened 2022-06-16
Rewrite HTTP 403 to 404 for index.crates.io

arlosi:
S3 returns HTTP 403 when an object is not found, which leads to confusing errors in Cargo if a crate does not exist. 

This change causes CloudFront to rewrite the HTTP 403 from S3 to HTTP 404, which makes cargo handle it appropriately.

### DIFF
--- a/terraform/crates-io/impl/cloudfront-index.tf
+++ b/terraform/crates-io/impl/cloudfront-index.tf
@@ -41,6 +41,14 @@ resource "aws_cloudfront_distribution" "index" {
     }
   }
 
+  // Rewrite 403 (access denied) to 404 (not found). S3 returns 403 when an
+  // object is not found, which leads to confusing errors in Cargo if a 
+  // crate does not exist.
+  custom_error_response {
+    error_code    = 403
+    response_code = 404
+  }
+
   origin {
     origin_id   = "main"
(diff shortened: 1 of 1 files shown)


Urgau:
From what I can see this StackOverflow [issue](https://stackoverflow.com/a/19038017), the 403 is return because there is a missing permission `s3:ListBucket` on the S3 bucket.

According to https://jamesthom.as/2021/02/aws-s3-iam-errors-with-missing-files-404-expected-403-returned/ adding something like this here should do the trick: https://github.com/rust-lang/simpleinfra/blob/d99cd2c95efcfbd38fe7c5f96ada1d02e9b9a890/terraform/crates-io/impl/s3-index.tf#L12-L25
```
{
    Effect = "Allow"
    Principal = {
          AWS = "*"
    }
    Action = "s3:ListBucket"
    Resource = "${aws_s3_bucket.index.arn}"
},
```

NOTE: I have no experience with Terraform nor AWS services I might be completely wrong but this feels like a proper solution.

Mark-Simulacrum:
I'm going to go ahead and close this; I've pushed a different version which enables List access on the bucket via CloudFront origin access identities.

arlosi:
Thanks! Sounds like the right solution.

PR #104, opened 2022-06-17
Allow heroku to delete objects from s3

carols10cents:
We have [an admin tool in crates.io](https://github.com/rust-lang/crates.io/blob/74a41e6fd2e86910484770a65224f628ec76d71d/src/admin/delete_crate.rs) that lets us remove crates in certain circumstances.

As part of the HTTP index work, [removing the crate from the HTTP index was added to this admin tool](https://github.com/rust-lang/crates.io/commit/acb38cb6a9eefc377c9bdde74b0289e3f9c1d797#diff-464e2ad5d7f91f317a36db0c511fb60c6fa4bf38c5a8e1530d4c17d5b527c833R51). This currently fails because the heroku AWS creds don't have s3 DeleteObject permissions.

If you don't want the heroku creds to have DeleteObject permissions, please close this. Instead, I'll remove that line from the admin tool and we'll clean up the HTTP index with individual credentials.

If I've done this incorrectly, please let me know what to fix and I'd be happy to!

### DIFF
--- a/terraform/crates-io/impl/iam.tf
+++ b/terraform/crates-io/impl/iam.tf
@@ -20,6 +20,7 @@ resource "aws_iam_policy" "static_write" {
           "s3:GetObjectAcl",
           "s3:PutObject",
           "s3:PutObjectAcl",
+          "s3:DeleteObject",
         ]
         Resource = [
           "${aws_s3_bucket.static.arn}/*",


Mark-Simulacrum:
Hm, this seems OK, but I'd like to see if we can avoid granting permission to delete object versions permanently (just "hide" objects by putting a delete marker in front of them). Not sure if that's possible yet, will need to look into the IAM rules we can add.

Mostly this is defense in depth, but it seems generally appropriate.

I think in the future we'll probably want to move to having admin tool users authenticate with their own set of AWS credentials, so we can have an audit trail for who is performing the delete (and maybe log a reason why).

carols10cents:
Doesn't the bucket have versioning turned on? Doesn't that take care of it?

Are you saying AWS has a "hidden" object feature? Do you have a link to docs on that? I haven't heard of it.

Or are you saying we should move the files to __deleted/crates/whatever.1.0.0.crate instead of deleting them? That's definitely not what happens now. What about when multiple people have tried spamming dictionary words with 0.1.0 versions, so that we have to delete /crates/aardvark.0.1.0.crate multiple times?

Mark-Simulacrum:
The bucket does have versioning turned on, but I believe DeleteObject (when called with the right parameters) can delete a version, not just the latest version. Deleting a specific version does actually delete that version permanently, I believe, not just adding a delete marker on top.

Mark-Simulacrum:
FWIW, I would be much more OK with this if we were only granting delete access to the index, not the static bucket -- it looks like the delete tool only deletes from the index at a quick skim?

pietroalbini:
> Hm, this seems OK, but I'd like to see if we can avoid granting permission to delete object versions permanently (just "hide" objects by putting a delete marker in front of them). Not sure if that's possible yet, will need to look into the IAM rules we can add.
>
> Mostly this is defense in depth, but it seems generally appropriate.

I can't find a way to scope the privileges with IAM, but at least for `crates-io` (static) the protection is already in place thanks to replication. Heroku only has access to the `crates-io` bucket, which is then replicated to the `crates-io-fallback` bucket, and S3 replication doesn't replicate permanent deletions (it replicates only delete markers). It's impossible to permanently delete something with only Heroku's credentials.

We should setup replication for the index bucket too before stabilizing the feature (to make the whole thing more resilient to outages), and that would remove the permanent deletion concern from the index too.

> FWIW, I would be much more OK with this if we were only granting delete access to the index, not the static bucket -- it looks like the delete tool only deletes from the index at a quick skim?

At least when I was in the crates.io team, allowing the tool to delete objects from static was also desirable, as it's a required step of deleting a crate or a version. The deletion would have to be done manually anyway, and it would've been awesome to have it automated.

Mark-Simulacrum:
OK, great -- I wasn't aware they don't get replicated. No problems moving ahead with this as-is, I mostly am concerned about accidental/unintentional data loss for this tooling, at least in the shape it is now. So I think we should be OK merging and deploying this; I'm not sure I'll get a chance today but either Pietro or I will do it soon I imagine.

pietroalbini:
[Link to the replication docs for future reference.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-what-is-isnot-replicated.html#replication-delete-op)

@carols10cents while I was at it, I also allowed Heroku to issue CloudFront invalidations for `index.crates.io` and `static.crates.io` (plus the staging domains). The former will allow to improve the index implementation, and the latter to automate yet another step of the crate deletion process. I also configured the `CLOUDFRONT_DISTRIBUTION_ID_INDEX` and `CLOUDFRONT_DISTRIBUTION_ID_STATIC` with the appropriate distribution IDs on both staging and production.

carols10cents:
Great, thank you!!

PR #110, opened 2022-06-26
init.sh: set username and email for git

fee1-dead:


ansible/roles/dev-desktop/files/scripts/init_git.py
```diff
@@ -0,0 +1,41 @@
...
+import re
+import subprocess
+import getpass
+
+for cfg in ["email", "name"]:
+    if subprocess.run(["git", "config", "--global", f"user.{cfg}"], stdout=subprocess.DEVNULL).returncode == 0:
+        print("git username and email already configured")
+        exit(0)
+
+team = requests.get("https://raw.githubusercontent.com/rust-lang/team/gh-pages/v1/teams/cloud-compute.json")
```
pietroalbini, line 12: This should use the endpoint for the permission, the `cloud-compute` team is temporary:

https://team-api.infra.rust-lang.org/v1/permissions/dev_desktop.json

ansible/roles/dev-desktop/files/scripts/init_git.py
```diff
@@ -0,0 +1,41 @@
...
+username = getpass.getuser()
+username = username.lstrip("gh-")
+
+print(f"found github name: {username}")
+
+for member in team.json()["members"]:
+    github = member["github"]
+    name = member["name"]
+    if github == username:
+        profile = f"https://raw.githubusercontent.com/rust-lang/team/master/people/{github}.toml"
```
pietroalbini, line 24: Hmm, rather than querying the file we should have a new team-api.infra.rust-lang.org endpoint to get the information we need.

pietroalbini (changes requested):

fee1-dead:
Updated and used the api. Tested on the cloud desktop.

ansible/roles/dev-desktop/files/scripts/init_git.py
```diff
@@ -0,0 +1,42 @@
...
+username = getpass.getuser()
+username = username.lstrip("gh-")
+
+print(f"found github name: {username}")
+
+for github_name in dev_desktop.json()["github_users"]:
+    if github_name == username:
+        if github_name not in people_map:
+            print("could not find a matching user in the people map!")
+            continue
```
jdno, line 28: I might be missing something, but why are we checking `dev_desktop.json` first? We already know that the `username` is the GitHub name, based on the conventions that are used to create the user accounts. And the user cannot run the script without being in that file. Instead of fetching `dev_desktop.json` and checking there first, can't we simply look for the user in `people.json`?

```python
if username not in people_map:
    print("could not find a matching user in the people map!")
    continue
```

jdno (approved):
Looks good, thanks @fee1-dead!

PR #117, opened 2022-08-27
Migrate sync-team + team repos to OIDC authentication

Mark-Simulacrum:
cc https://github.com/rust-lang/sync-team/pull/7, https://github.com/rust-lang/team/pull/813

### DIFF
--- a/terraform/team-repo/ci.tf
+++ b/terraform/team-repo/ci.tf
@@ -7,39 +7,42 @@ module "ecr" {
   name   = "sync-team"
 }
 
-// IAM User used rust-lang/sync-team's CI to push the built images on ECR.
+// IAM role used by rust-lang/sync-team's CI to push the built images on ECR.
 
-module "iam_ci_sync_team" {
-  source = "../shared/modules/gha-iam-user"
+module "ci_sync_team" {
+  source = "../shared/modules/gha-oidc-role"
   org    = "rust-lang"
   repo   = "sync-team"
+  branch = "master"
 }
 
-resource "aws_iam_user_policy_attachment" "ci_sync_team_pull" {
-  user       = module.iam_ci_sync_team.user_name
+resource "aws_iam_role_policy_attachment" "ci_sync_team_pull" {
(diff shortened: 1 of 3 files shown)


terraform/team-repo/ci.tf
```diff
@@ -7,50 +7,82 @@ module "ecr" {
...
 }
 
-// IAM User used rust-lang/sync-team's CI to push the built images on ECR.
+// IAM role used by rust-lang/sync-team's CI to push the built images on ECR.
 
-module "iam_ci_sync_team" {
-  source = "../shared/modules/gha-iam-user"
-  org    = "rust-lang"
-  repo   = "sync-team"
+resource "aws_iam_role" "ci_sync_team" {
```
pietroalbini, line 12: Code looks good! In general, the assume role policy can be a bit finicky to write, and I'm wondering if it's worth creating a module like:

```terraform
module "ci_sync_team_role" {
  source = "../shared/modules/gha-oidc-role"
  name = "ci--rust-lang-sync-team"
  repo = "rust-lang/sync-team"
  branch = "master"
  policy = jsonencode({ ... })
}
```

pietroalbini (approved):

terraform/team-repo/ci.tf
Mark-Simulacrum: Yeah, I thought about it. I'm not sold on a module making this much better (my main annoyance with modules is having to reapply all over the place after editing them) and it seems like we're currently copy pasting anyway.

pietroalbini: Hmm yeah that part is annoying, even though for this I don't really expect the module to change. The only change I can see is adding other options other than `branch`, and those shouldn't affect existing impls of the module. I think with how low risk of this module changing is, I'd prefer to switch to one so that we don't risk messing up the assume policy.

Mark-Simulacrum: Yeah, seems ok. I'll make that update when I get a chance and migrate existing module and the PRs here over.

Issue #113, opened 2022-07-28
Opening up UDP ports on Cloud Compute Servers


JoelMarcey:
I got a request asking to open up UDP ports 60000-61000 so that mosh can work correctly. Is this something we can do? Any high security issues we should consider?

Also, tangentially, I was asked if we could set `kernel.yama.ptrace_scope=0` so that ptrace can attach to existing processes. This is probably not as high of an ask as the UDP ports.

PR #118, opened 2022-09-16
Install Mosh on dev desktops

jdno:
Mosh has been added to the dev desktops. The ports that it uses have been allowed in the security group on AWS as well as the local firewall on the machine.

Requested in https://github.com/rust-lang/simpleinfra/issues/113


pietroalbini (approved):
Looks great, thanks!

PR #119, opened 2022-09-22
Allow users to debug their processes

jdno:
Ubuntu restricts ptrace by default, and only allows a process to debug its own child processes. This is limiting the ability of developers to use the dev desktops to analyze and debug their work.

The default scope for ptrace has been changed to `0`, allowing users to debug their own processes. This carries with it the risk that malicious code can interact with any process that a user created. We deem this to be acceptable in this specific use case, as the dev desktops are intended to be used only for the development of Rust and have, by design, no user data or credentials that could be extracted.

Requested in #113

### DIFF
--- a/ansible/roles/dev-desktop/tasks/main.yml
+++ b/ansible/roles/dev-desktop/tasks/main.yml
@@ -26,7 +26,8 @@
     path: /etc/sysctl.d/10-ptrace.conf
     regexp: '^kernel.yama.ptrace_scope = [\d]$'
     replace: 'kernel.yama.ptrace_scope = 0'
-  register: ptrace_scope
+  notify:
+    - reboot-machine
 
 - name: Configure update script service
   copy:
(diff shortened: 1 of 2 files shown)


ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -172,3 +179,18 @@
     src: sudoers
     dest: /etc/sudoers.d/dev-desktop
     mode: 0440
+
+- name: Reboot to apply sysctl changes
```
jdno, line 183: I am not sure if rebooting as part of the role can cause issues with existing workflows. A reboot is definitely required after changing `kernel.yama.ptrace_scope`, but maybe there is a better way to ensure that it happens.

(This will also come up with the quotas, since they require a restart as well to enable the feature.)

pietroalbini: Can we do the reboot with a handler? That way if multiple steps require a reboot we only do it once.

pietroalbini: There is also a [reboot module](https://docs.ansible.com/ansible/2.7/modules/reboot_module.html).

jdno: Oh my. The module is so nice. Role has been updated to use a handler and the module. :+1:

pietroalbini (approved):

PR #120, opened 2022-09-23
Set a filesystem quota for dev desktop users

jdno:
The Ansible role for dev desktops has been expanded to enable filesystem quotas for users. The actual quotas are set when a user is created by the team login feature.

### DIFF
--- a/ansible/roles/dev-desktop/files/team_login/src/main.rs
+++ b/ansible/roles/dev-desktop/files/team_login/src/main.rs
@@ -90,6 +90,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .success(),
             "failed to set the default shell"
         );
+
+        // Set a user quota
+        assert!(
+            cmd("setquota", &["-u", &username, "50G", "51G", "0", "0", "/"])?
+                .status
+                .success(),
+            "failed to set a user quota"
+        );
     }
     // Delete all keys for users that weren't on the list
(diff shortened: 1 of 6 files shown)


jdno:
I've tested this against a local virtual machine, but the quota system is fiddly and it's probably best to run this against the staging instance before merging.

ansible/roles/dev-desktop/files/team_login/src/main.rs
```diff
@@ -90,6 +90,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .success(),
             "failed to set the default shell"
         );
+
+        // Set a user quota
+        assert!(
+            cmd("setquota", &["-u", &username, "50G", "51G", "0", "0", "/"])?
```
pietroalbini, line 96: Can we make this configurable from the playbook?

jdno: We sure can! I went through two iterations, first setting the quota as an environment variable and then passing it as an argument to the CLI. I like the second option much more, since it makes the dependency explicit. The downside is that I needed/wanted to pull in `clap` to parse the arguments.

jdno:
What's the preferred strategy for both resolving the merge conflict and then merging the pull request?

pietroalbini (approved):
To solve the conflicts I'd recommend to rebase on top of `master` locally.

PR #124, opened 2022-10-04
Enable perf on dev desktops

jdno:
The package for the `perf` utility has been added to the configuration for the dev desktops. The kernel setting that enables perf has been set to 2, which should provide a good mix of functionality and safety.

### DIFF
--- a/ansible/roles/dev-desktop/defaults/main.yml
+++ b/ansible/roles/dev-desktop/defaults/main.yml
@@ -6,10 +6,5 @@ allow_ssh_extra_groups: "dev-desktop-allow-ssh"
 # Filesystem quota per user in GB
 vars_user_quota_gb: 50
 
-# For Ubuntu hosts, change this in a host variable to:
-# "linux-modules-extra-virtual"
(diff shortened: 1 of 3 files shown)


ansible/roles/dev-desktop/defaults/main.yml
```diff
@@ -9,3 +9,7 @@ vars_user_quota_gb: 50
 # For Ubuntu hosts, change this in a host variable to:
 # "linux-modules-extra-virtual"
 vars_user_quota_linux_package: "linux-modules-extra-aws"
+
+# For Ubuntu hosts, change this in a host variable to:
+# "linux-tools-common"
+vars_perf_linux_package: "linux-tools-aws"
```
pietroalbini, line 15: Can we change both variables to `vars_linux_kernel_flavor` that can be "aws" or "virtual"? `linux-tools-virtual` exists as well, so it should be just possible to suffix the package names with the kernel flavor.

pietroalbini (changes requested):

pietroalbini (approved):

PR #126, opened 2022-10-05
Install kernel-specific packages

jdno:
The package required to enable user quotas installs kernel modules, which will use the latest version of the kernel by default. This can cause issues, since the system might be running an older version. In this scenario, modprobe will fail to find the modules and return an error.

Instead of installing the latest version of the package, we are now discovering the current version of the kernel and installing the matching package.


jdno:
Noticed an issue while testing on staging where enabling quotas requires the current version of the kernel. But after restarting the server at the end of the playbook, the latest version is required. So I guess we have to install both, just to be safe.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -11,7 +15,9 @@
       - python3-cryptography
       - quota
       - "linux-modules-extra-{{ vars_linux_kernel_flavor }}"
```
pietroalbini, line 17: Can we also calculate the kernel flavor from `uname -r`? it should be the last item when splitting by `-`.

pietroalbini (approved):

PR #134, opened 2022-10-12
Update dependencies of cancel-outdated-builds

JohnTitor:
Updated deps as it's too old. It'd be great if we could somehow ensure the behavior isn't changed (maybe merging into another branch and testing there?).

Signed-off-by: Yuki Okushi <jtitor@2k36.org>


rylev:
I suppose perhaps we can just set up a GitHub repo that uses this in its CI and see if there are any issues when it runs.

rylev:
Closing in favor of #138

PR #135, opened 2022-10-12
Upgrade some actions to Node.js 16

JohnTitor:
GHA now shows warnings if we use Node.js 12 on actions: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
That adds a bunch of warnings on the libc repo: https://github.com/rust-lang/libc/actions/runs/3236239538
This updates actions that use Node.js to 16 to get rid of them.
Signed-off-by: Yuki Okushi <jtitor@2k36.org>


rylev (approved):
This looks good. I guess we'll want to monitor repos that use this action carefully after we merge it.

PR #136, opened 2022-10-18
Enable response header policy for staging.crates.io

jdno:
A response header policy has been added to the CloudFront distribution for crates.io to enable the `includeSubdomains` directive in the strict transport security header.

The policy is hidden behind a variable right now to allow us to test it on staging first. When we decide to roll it out to production, we can remove the variable again.

### DIFF
--- a/terraform/crates-io/impl/cloudfront-webapp.tf
+++ b/terraform/crates-io/impl/cloudfront-webapp.tf
@@ -25,6 +25,8 @@ resource "aws_cloudfront_distribution" "webapp" {
     min_ttl     = 0
     max_ttl     = 31536000 // 1 year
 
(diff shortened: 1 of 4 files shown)


terraform/crates-io/impl/cloudfront-webapp.tf
```diff
@@ -125,3 +127,20 @@ resource "aws_route53_record" "webapp_apex" {
     evaluate_target_health = false
   }
 }
+
+resource "aws_cloudfront_response_headers_policy" "webapp" {
```
rylev, line 131: Would it make sense to link to the [issue](https://github.com/rust-lang/crates.io/issues/5332) in a comment so that it's very obvious *why* we're doing this?

jdno: Great idea! I've added a comment linking to the issue. :+1:

rylev (approved):

PR #137, opened 2022-10-20
Fix node-fetch import

rylev:
Without this fix, we get a `fetch is not a function` error when running. 

This needs to merge first and then we can merge #134 and #135


jdno (approved):
👍

PR #140, opened 2022-10-22
dev-desktop: add nodejs package for running rustdoc-gui tests

fee1-dead:


jdno (changes requested):
The `nodejs` package installs Node 12.22.9, which is officially end-of-life. If we want to install Node globally, I'd much rather use [NodeSource](https://github.com/nodesource/distributions/blob/master/README.md) to set up Node 16 or 18.

But personally, I'd prefer using [nvm](https://github.com/nvm-sh/nvm) so that users can install the specific Node version(s) that they need. Would that work?

Mark-Simulacrum:
Most rustc contributors aren't actually going to run node for anything other than the rustdoc-gui tests, so I'm not sure a per-user installation (via nvm) makes sense here.

I think Ubuntu wants us to use snap (i.e., https://snapcraft.io/node) to install node, which I think is a global installation and gets us latest stable?

Mark-Simulacrum:
(But I'm totally fine and in fact prefer the NodeSource option, I think).

jdno:
I don't have a strong opinion on anything other than avoiding Node 12. 😅 Either snap or NodeSource is fine with me.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -186,6 +186,11 @@
       - gcc-multilib # for 32 bit cross compiles of libstd and tests
     state: present
 
+- name: Install Nodejs
+  shell: |
+    curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
+    apt-get install -y nodejs
```
jdno, line 192: Let's add a quick check to see if Node is already installed and use the `apt` module to install the package:

```suggestion
- name: Check if Node is installed
  command: node --version
  register: node_version
  changed_when: false
  failed_when: node_version.rc != 0 and node_version.rc != 2

- name: Set up NodeSource repository
  shell: |
    curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
  when: node_version.rc == 2

- name: Install Node
  apt:
    name: nodejs
    state: present
    update_cache: yes
```

jdno (changes requested):
Thanks for your efforts, @fee1-dead! I've added a check and split the task to use the `apt `module (see suggestion). Also tested the playbook on a fresh machine and checked that Node works as expected.

jdno (approved):

jdno:
I've just deployed to the new dev desktops, so Node should be set up there. If you find any issues, @fee1-dead, please let me know in an issue or on Zulip.

PR #141, opened 2022-10-24
Enable response header policy for crates.io

jdno:
The new HSTS header that sets the `includeSubdomains` directive has been rolled out to production, after having been tested on staging [^1].

[^1]: https://github.com/rust-lang/simpleinfra/pull/136


jdno:
Yes, it will be! I want to leave the variable in place for a day or so in case any issues pop up.

PR #148, opened 2022-11-10
Manage team members on Fastly

jdno:
A new Terraform configuration has been added to manage team members on Fastly. Members of the infra-team have been invited to the account, and we'll add more teams as we start using the service.

### DIFF
--- a/terraform/team-members-fastly/_terraform.tf
+++ b/terraform/team-members-fastly/_terraform.tf
@@ -19,11 +19,4 @@ terraform {
   }
 }
 
-provider "fastly" {
-  api_key = var.fastly_api_key
-}
-
-variable "fastly_api_key" {
-  description = "Personal API token for Fastly"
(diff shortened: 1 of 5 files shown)


terraform/team-members-fastly/_terraform.tf
```diff
@@ -0,0 +1,29 @@
...
+}
+
+provider "fastly" {
+  api_key = var.fastly_api_key
+}
+
+variable "fastly_api_key" {
+  description = "Personal API token for Fastly"
+  type        = string
+}
```
pietroalbini, line 29: Can we use the environment variable instead, and remove the tfvars from gitignore? https://registry.terraform.io/providers/fastly/fastly/latest/docs#environment-variables

pietroalbini (changes requested):

pietroalbini (approved):

PR #154, opened 2022-11-18
Rust root organization

rylev:
This adds terraform configuration for the newly created Rust organization.

I have only added @jdno as a user so far as I want to confirm this is the right direction before adding others.

`terraform plan` shows minimal changes will actually be applied when this is run:
* Descriptions for the identitystore groups
* @jdno's email no longer being marked as a work email

Once we're happy with this, I will add the other users and run `terraform apply`.


terragrunt/aws/root/aws-organization/terragrunt.hcl
```diff
@@ -0,0 +1,8 @@
+terraform {
+  source = "../../../modules//aws-organization"
```
Mark-Simulacrum, line 2:
```suggestion
  source = "../../../modules/aws-organization"
```

Mark-Simulacrum (approved):

terragrunt/aws/root/aws-organization/terragrunt.hcl
pietroalbini: The double slash is actually intentional. From the terragrunt docs:

> Terragrunt will download all the code in the repo (i.e. the part before the double-slash `//`) so that relative paths work correctly between modules in that repo.

Mark-Simulacrum: Oh. We should have a comment, then.

jdno: This is actually a feature of Terraform: https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories

pietroalbini: @Mark-Simulacrum the `//` will be in every Terragrunt configuration file, would you expect a comment in each file?

Mark-Simulacrum: I guess I know about it now so maybe not so important. I think having a link in each file isn't terrible though.

pietroalbini (approved):

PR #155, opened 2022-11-21
Create Fastly service for static.crates.io

jdno:
This is a first pass at a Fastly service for `static.crates.io` (tracking issue https://github.com/rust-lang/infra-team/issues/62).


terraform/crates-io/impl/fastly-static.tf
```diff
@@ -0,0 +1,44 @@
+resource "fastly_service_vcl" "static" {
+  name = var.static_domain_name
+
+  domain {
+    name = var.static_domain_name
+  }
+
+  backend {
```
jdno, line 8: We are using an `origin_group` on CloudFront with two S3 buckets and a failover strategy based on HTTP `5xx` responses. Achieving the same setup is slightly involved on Fastly:

- The `fastly_service_vcl` supports [failovers](https://developer.fastly.com/learning/concepts/failover/) through _automatic load-balancing_ or a _fallback director_. The latter is [not supported yet](https://github.com/fastly/terraform-provider-fastly/issues/436) in Terraform, and the first requires us to set up health checks against S3.
- We can probably write some custom VCL to replicate the functionality.
- The `fastly_service_compute` allows us to deploy a [Compute@Edge](https://github.com/fastly/compute-starter-kit-rust-static-content/blob/main/src/main.rs) function, which can mirror CloudFront's behavior. Requires building a local WASM module.

The benefit of `compute` over `vcl` seems to be that we have full control over the request cycle in a Rust runtime. The downside is that we are now relying on our own potentially buggy code instead of a provided Terraform module. If we do custom stuff, I'd still rather write some Rust than VCL.

Mark-Simulacrum: I feel like I commented somewhere but can't find it -- one alternative is to "just" have Fastly hit cloudfront as its origin, right? We can add a CNAME to cloudfront (e.g., cloudfront.static.crates.io) which Fastly hits for its origin.

terraform/crates-io/impl/fastly-static.tf
```diff
@@ -0,0 +1,44 @@
+resource "fastly_service_vcl" "static" {
```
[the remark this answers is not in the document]
Mark-Simulacrum: I think long-term the answer here is probably two different terraform state files (and variable-based config), one for staging and one for production, and we pick one based on a tfvars file or similar.

We hit this in ~every area we have a dev vs. prod distinction and we need to invest in fixing it more generally without hacks like for_each maps or similar.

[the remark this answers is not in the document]
jdno: For the new AWS organization with SSO, @pietroalbini suggested using [terragrunt](https://terragrunt.gruntwork.io/).

The workaround for now can be to run

```shell
terraform plan -target=module.staging
```

At least as long as the changes are in a separate branch.

terraform/crates-io/impl/fastly-static.tf
pietroalbini: Another problem would be dealing with invalidations. If both CDNs are independent we can issue invalidations to both, while if Fastly points to CloudFront we would first need to invalidate CloudFront, wait for the invalidation to be finished, and then invalidate Fastly. That sounds like a mess to implement.

terraform/crates-io/compute-static/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+/// queried. If the response indicates a server issue (status code >= 500), the request is sent to
+/// a fallback bucket in a different geographical region.
+fn send_request_to_s3(config: &Config, request: &Request) -> Result<Response, Error> {
+    let primary_request = request.clone_without_body();
+
+    let mut response = primary_request.send(&config.primary_host)?;
+    let status_code = response.get_status().as_u16();
+
+    if status_code >= 500 {
+        warn!(
```
jdno, line 50: The setup for [logging](https://developer.fastly.com/learning/compute/rust/#logging) is not complete yet, and we are not [streaming logs](https://docs.fastly.com/en/guides/about-fastlys-realtime-log-streaming-features). So this warning will not show up anywhere.

terraform/crates-io/impl/fastly-static.tf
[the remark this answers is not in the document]
pietroalbini: I'd split crates.io into terragrunt ASAP and target this PR on there.

terraform/crates-io/impl/fastly-static.tf
pietroalbini: If possible I'd prefer not to have a Rust binary handling this. It makes deploying changes to the CDN configuration more complex, and has a higher chance of bugs.

> and the first requires us to set up health checks against S3.

While it wouldn't be exactly the same behavior as CloudFront, setting up a health check would be trivial: we can create an empty[^1] `health` file into the S3 bucket and have Fastly use that as the health check. It would still cover the reason why the fallback is there (S3 is down in that region), even though it's not exactly the same mechanism as CloudFront.

[^1]: Well, in the content of the health file I'd put something like "needed by the Fastly CDN for the health check" so that future us will know why that file exists.

terraform/crates-io/impl/fastly-static.tf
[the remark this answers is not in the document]
jdno: Yes, that's absolutely the goal!

terraform/crates-io/impl/fastly-static.tf
jdno: Even with a health check, we will not be able to set up a primary/fallback configuration with Fastly. Fastly can only load balance between two buckets, and retry the other if the first request fails. Which is not a big issue in my opinion, but something to note. (See the remark about the _fallback director_ [not being supported](https://github.com/fastly/terraform-provider-fastly/issues/436) in Terraform.)

Personally, I don't have an issue using a compute function here. I think that we will need a binary for the Rust releases, since that does a lot of rewriting that cannot be expressed easily in VCL (as far as I can tell). So sooner or later, we have to learn how to manage it anyways.

But I'm fine leaving that discussion for later, and configuring `static.crates.io` with VCL for now.

jdno:
Just a heads-up that I'll be setting up terragrunt for the `crates-io` module, and then I will rebase this PR.

jdno:
I've rebased the pull request on a very recent commit in `master` and moved the resources into the new `terragrunt` structure.

jdno:
The following things still need to be done:

- Set up weighted routing for DNS records
  - Rename the domain name of the CloudFront distribution (e.g. `cloudfront-${var.static_domain_name}`)
  - Add a variable to configure the weights
  - Create a DNS record for `var.static_domain_name` for each CDN
- Set TTL explicitly
- Set up log streaming
  - Figure out where to send the logs
  - Configure the log destination
- Add some tests (if it's possible to unit test the function)

My preference would be to merge this pull request and then work through these items in separate branches. This pull request is long enough as it is, and it'll be easier to review the changes one-by-one.

jdno:
This feels like a good place to stop and merge.

The configuration only sets up Fastly right now, which will make it possible to deploy it to production as well and test there without touching any existing resources.

All changes beyond this point will interact with the CloudFront distribution or with DNS, which we probably want to review and test separately.

terragrunt/modules/crates-io/fastly-static.tf
```diff
@@ -0,0 +1,123 @@
+# The Fastly service must be deployed in two steps, since some resources depend
```
rylev, line 1: I'm surprised by this - isn't terraform supposed to be able to handle dependencies like this?

jdno: Yes and no.

The problem here is that the newly created resources are used as keys in a `for_each` statement:

```text
Error: Invalid for_each argument

  on fastly-index.tf line 74, in resource "aws_route53_record" "index_tls_validation":
  74:   for_each = {
  75:     for challenge in fastly_tls_subscription.index.managed_dns_challenges :
  76:     trimprefix(challenge.record_name, "_acme-challenge.") => challenge
  77:   }
    ├────────────────
    │ fastly_tls_subscription.index.managed_dns_challenges is a set of object, known only after apply

The "for_each" map includes keys derived from resource attributes that
cannot be determined until apply, and so Terraform cannot determine the
full set of keys that will identify the instances of this resource.

When working with unknown values in for_each, it's better to define the map
keys statically in your configuration and place apply-time results only in
the map values.

Alternatively, you could use the -target planning option to first apply
only the resources that the for_each value depends on, and then apply a
second time to fully converge.
```

The same problem exists when rewriting this block to use `count`:

```text
Error: Invalid count argument

  on fastly-index.tf line 72, in resource "aws_route53_record" "index_tls_validation":
  72:   count = length(fastly_tls_subscription.index.managed_dns_challenges)

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be
created. To work around this, use the -target argument to first apply only
the resources that the count depends on.
```

The only potential workaround I've found is splitting the configuration into two modules, but those would still need to be applied after each other. We can do that with dependencies in Terragrunt, but that would mean we need a state for the `fastly_tls_subscription` and a state for the validation.

```text
crates-io-staging
├── crates-io
├── fastly-tls-subscription
└── fastly-tls-validation
```

One could then apply all changes by running `terragrunt run-all apply` from the `crates-io-staging` directory, which will resolve the dependencies and apply the changes in the right order.

Thinking more about this, I kinda like this approach. It makes the dependency explicit in the tooling, and not in the documentation. What do you think?

jdno: I've put together a quick proof-of-concept here: https://github.com/jdno/rust-simpleinfra/commit/28bb026d0844ae45920c2738d7b0ac3db6738821 Haven't tried actually running it, though.

rylev: Without too much experience here, I can't have a strong opinion, but the suggestion feels better than commenting out code.

terragrunt/modules/crates-io/compute-static/Cargo.lock
```diff
@@ -0,0 +1,394 @@
+# This file is automatically @generated by Cargo.
```
pietroalbini, line 1: Is this lockfile actually used? The dependencies should be included in the one at the root of the repo due to the workspace.

terragrunt/modules/crates-io/compute-static/README.md
```diff
@@ -0,0 +1,23 @@
...
+cd compute-static
+fastly compute build
+```
+
+Export an API token for Fastly and then run Terraform:
+
+```shell
+export FASTLY_API_KEY=""
+terraform plan
+```
```
pietroalbini, line 23: I'm worried that you have to remember to run `fastly compute build` to get the right configuration, and if you don't changes will be silently not applied. Can we use the [`external` data source](https://registry.terraform.io/providers/hashicorp/external/latest/docs/data-sources/data_source) to invoke the compilation as part of planning?

pietroalbini (changes requested):

terragrunt/modules/crates-io/fastly-static.tf
pietroalbini: I'm a bit confused by this, how's [AWS validation working then?](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation) It also generates the DNS records dynamically.

pietroalbini:
> * Set up log streaming
>   * Figure out where to send the logs
>   * Configure the log destination

Note that we do not collect logs from CloudFront either as of today.

terragrunt/modules/crates-io/compute-static/Cargo.lock
jdno: It wasn't until I added the external data source. But since Terragrunt copies the Rust crate into a temporary build environment, we cannot provide a stable path to include the crate in the workspace.

terragrunt/modules/crates-io/compute-static/README.md
jdno: Done. I had to add a shell scripts that runs `fastly compute build` and returns a valid JSON object, though.

jdno:
Log streaming might be interesting for debugging our functions. Not so much for request logs.

terragrunt/modules/crates-io/fastly-static.tf
jdno: Here is the relevant part of a new `aws_acm_certificate` in a Terraform plan:

```
resource "aws_acm_certificate" "cert" {
  domain_validation_options = [
    {
      domain_name           = "test.jdno.dev"
      resource_record_name  = (known after apply)
      resource_record_type  = (known after apply)
      resource_record_value = (known after apply)
    },
  ]
```

And here is the `fastly_tls_subscription`:

```
resource "fastly_tls_subscription" "subscription" {
   managed_dns_challenges  = (known after apply)
```

The reason why AWS works is that the number of `domain_validation_options` with their respective `domain_name`, i.e. the key for the `for_each` statement, is known when planning.

We can create a feature request to get this added to [terraform-provider-fastly](https://github.com/fastly/terraform-provider-fastly), but in the meantime we are stuck with either running the _initial_ apply twice or using dependencies.

Opinions, @pietroalbini and @rylev?

jdno: I've created a feature request for the Fastly provider: https://github.com/fastly/terraform-provider-fastly/issues/627

pietroalbini: Ooh ok, your explaination on the issue (unfortunately) makes a lot of sense. Ugh. I'd go with the two applies for now. Also, can you flag the issue with our Fastly contact to see if it can be prioritized on their end?

pietroalbini (approved):

terragrunt/modules/crates-io/fastly-static.tf
jdno: I forwarded both issues to our contact and will post an update in the team meeting when I know more about the priorities.

PR #157, opened 2022-11-25
Deploy dev desktops to Azure

jdno:
Two dev desktops have been deployed to Azure. They have a similar configuration to the AWS instances (32 vCPU, 64GB memory), but use the x86 architecture.


jdno:
I'll update the Ansible hosts when we decommission the original instance. Is 1 week notice enough for users to migrate?

terraform/dev-desktops/_terraform.tf
```diff
@@ -45,3 +45,8 @@ provider "aws" {
   region  = "us-east-1"
 }
 
+provider "azurerm" {
+  features {}
```
Mark-Simulacrum, line 49: What does auth for this look like locally? I sort of think maybe it makes sense to split into separate directories - it can be annoying to need credentials for something you're not changing.

I wonder if we could pivot off of AWS credentials to fetch temporary azure credentials automatically.

jdno: That's a good point. Happy to split the module into `dev-desktops-aws` and `dev-desktops-azure`.

We haven't created a `team-members-azure` module yet to manage user accounts on Azure. With the limited knowledge that I have about Azure, it seems that we'd need to set up Azure AD and provision accounts there. That seemed a bit much just for spawning two virtual machines, so I've used the root account and the Azure CLI to authenticate.

jdno:
I moved the Azure module into terragrunt. Doesn't solve the problem of needing both AWS and Azure to run it, though. And we'll end up with a slightly weird setup, where some desktops are in `aws/legacy/dev-desktops` and others in `azure/microsoft-azure-sponsorship/dev-desktops`.

terragrunt/modules/dev-desktops-azure/firewall.tf
```diff
@@ -0,0 +1,74 @@
+resource "azurerm_network_security_group" "dev_desktops" {
```
rylev, line 1: Nit: I think some of these modules could use some comments with a short explainer about what they're doing.

PR #160, opened 2022-11-30
Add the rest of the AWS org users

rylev:
This adds the remaining users to the aws organization. It also refactors so that the users are not a local but rather are passed to the module. 

Note: I've already added these users to the terraform state.


jdno (approved):
The user part looks good, the groups I'm not so sure about. When the number of groups grows, we probably want to manage them in a single place as well. But let's tackle this in another PR.

PR #161, opened 2022-12-01
Account assignment

rylev:
This adds group to account assignments so that the existing groups (infra and infra-admins) have access to the 3 existing accounts (admin, docs-rs-staging, and dev-desktops-prod).

### DIFF
--- a/terragrunt/modules/aws-organization/groups.tf
+++ b/terragrunt/modules/aws-organization/groups.tf
@@ -5,6 +5,8 @@ locals {
   instance_arn      = data.aws_ssoadmin_instances.rust.arns[0]
 }
 
+# The various user groups
+
 resource "aws_identitystore_group" "infra-admins" {
   identity_store_id = local.identity_store_id
 
@@ -18,6 +20,8 @@ resource "aws_identitystore_group" "infra" {
   display_name = "infra"
   description  = "The infrastructure team"
 }
+
+# The different permission sets a group may have assigned to it
 
 resource "aws_ssoadmin_permission_set" "administrator_access" {
   instance_arn = local.instance_arn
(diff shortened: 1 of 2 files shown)


jdno (commented):
I am wondering if we are not making things to difficult for us. Can we reduce the complexity of our setup by putting a few conventions in place? For example:

- `infra-admins` is by default added to every account.
- For every account, we create a `${account_name}-admin` and `${account_name}-read" group with appropriate permissions sets.

Then we just have to manage users and can generate the assignments automatically from a list of account names. Because I expect that each account will only have 2-3 groups anyways: `infra-admins`, a group for maintainers, and maybe a read-only group.

This does assume that the "admin" access for all accounts looks the same, i.e. that we don't limit `docs-rs-staging` to just ECS and RDS, and `dev-desktop-prod` to just `EC2`. Which raises the bigger question on how we want to protect the accounts against mistakes/malice in the first place. `AdministratorAccess` might be a bit broad, and we probably want/need to create budget limits. The specifics of this are something we should discuss outside of this PR, however, we need to decide whether we want to apply generic admin privileges to all accounts.

jdno:
If we want to keep the flexibility of the current approach, I'm fine merging this to unblock other work and iterate in new pull requests.

jdno:
Also curious what @pietroalbini thinks...

pietroalbini:
I think we still want some of this flexibility, even though in the end I'd expect the configuration to just be:

```terraform
accounts = [
  {
    name = "docs-rs-staging"
    admin = ["infra-admins", "infra", "docs-rs"]
    # view = [], automatically adding admin to this
  },
  {
    name = "docs-rs-prod",
    admin = ["infra-admins"]
    view = ["infra", "docs-rs"] # automatically adding admin to this
  },
]
```

We can iterate on that later though (I might prepare a PR this weekend if it'd be helpful).

rylev:
I think I'd like to keep the flexibility. I like @pietroalbini's idea of abstracting over things with "admin" and "view" groupings. I can work on that (still trying to get more and more comfortable with HCL). But I think we should merge this first.

jdno (approved):
Ok, in that case I'm fine with merging this and iterating in other PRs on the remaining questions. :+1:

Thanks, @rylev! <3

PR #163, opened 2022-12-02
Reorganize configuration in terragrunt

jdno:
The configuration for terragrunt was previously organized by cloud providers. But every module relies on one way or another on AWS, which raised questions about this design (see Zulip[^1]).

The decision was made to organize the configuration by AWS accounts. Each account stores at least the state files for its configuration, and potentially has other shared resources such as DNS or S3 as well. Resources on other cloud providers can be emphasized by naming their module accordingly.

```text
accounts
├── dev-desktops-prod
│   ├── aws-eu-central-1
│   ├── aws-us-east-1
│   ├── azure-westeurope
│   └── azure-westus2
└── root
```

[^1]: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/terragrunt.20organization


rylev (approved):
We [discussed this a bit](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/terragrunt.20organization/near/313489911). I'm very much in favor of this change, but have some concerns about the name of directory. Since that can easily be changed later, I think we should merge.

jdno:
I agree that the name is not ideal yet. Let's have that discussion in Zulip, and create another PR once we found something better. :+1:

PR #165, opened 2022-12-08
Migrate crates.io into terragrunt

jdno:
The Terraform module for crates-io has been copied into the terragrunt modules folder, and slight adjustments have been made to adapt the module to terragrunt:

  - The certificate is now an external dependency instead of a submodule. The certificate ARN is passed as a variable.
  - The `inventories_bucket_arn` is now set as a `local` and not as a `variable`. This was necessary, since the remote state cannot be queried from the `terragrunt.hcl` configuration file.

The rest of the configuration remains unchanged.


pietroalbini (approved):

terragrunt/modules/crates-io/s3-static.tf
```diff
@@ -0,0 +1,142 @@
...
+  destination {
+    bucket {
+      bucket_arn = local.inventories_bucket_arn
+      prefix     = aws_s3_bucket.static.id
+      format     = "CSV"
+    }
+  }
+}
+
+resource "aws_s3_bucket" "fallback" {
```
rylev, line 113: Would be good to have a very short comment on the architecture here that requires the fallback bucket.

terragrunt/modules/crates-io/cloudfront-webapp.tf
```diff
@@ -0,0 +1,148 @@
...
+# Set strict-transport-security headers for crates.io and its subdomains
+# See https://github.com/rust-lang/crates.io/issues/5332 for details
+resource "aws_cloudfront_response_headers_policy" "webapp" {
+  count = var.strict_security_headers ? 1 : 0
+
+  name = replace(var.webapp_domain_name, ".", "-")
+
+  security_headers_config {
+    strict_transport_security {
+      access_control_max_age_sec = 31536000
```
rylev, line 140: Nit: comment that this is 1 day

terragrunt/modules/crates-io/iam.tf
```diff
@@ -0,0 +1,121 @@
+// This file configures the IAM users, policies and roles for crates.io
+
+resource "aws_iam_user" "heroku" {
+  name = "${var.iam_prefix}--heroku"
```
rylev, line 4: Do we need the prefix as a variable? Can we just turn the domain name into the prefix like we do with the `aws_cloudfront_response_headers_policy`?

terragrunt/modules/crates-io/iam.tf
```diff
@@ -0,0 +1,121 @@
...
+}
+
+resource "aws_iam_user_policy" "heroku" {
+  name = "inline"
+  user = aws_iam_user.heroku.name
```
rylev, line 9: Would be nice to comment on the invalidation architecture.

rylev (commented):
Looks good. I know this is predominantly a copy/paste job so no need to block on my comments, but I do think things would benefit from more comments. No knowing the crates.io architecture well, I found it hard to always know how exactly some of the pieces fit together. While I don't think we need a full dissertation on the architecture, some light commenting would go a long way.

terragrunt/accounts/legacy/crates-io-prod/deployed-ref
```diff
@@ -0,0 +1 @@
+25e405013655cafcf1eb6c2ce2423914a0df79eb
```
jdno, line 1: This is now referencing a commit in the same branch. I think we should update this to the merge commit once the PR has been merged into `master`.

jdno:
@rylev Let's have a chat about documentation for Terraform/terragrunt outside of this pull request. I agree that we should provide a short description of each variable, and more complex functions inside `locals` should probably be documented as well. But on the rest, I'd much rather have a `README` in each module that gives an high-level overview than comments on each resource.

jdno:
Having said that, I definitely agree that we need to document more! 🙈

pietroalbini (approved):
I would move the acm certificate to be a module inside `crates-io` again, but we can do so after this PR is merged (reasoning for that is, the certificate is not a separate component but just a bigger "resource" needed by the crates-io module, and the duplication of the domain names is annoying).

terragrunt/modules/acm-certificate/variables.tf
```diff
@@ -0,0 +1,4 @@
+variable "domains" {
+  description = "List of domain names included in the certificate"
```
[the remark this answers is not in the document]
pietroalbini: There is no distinction between primary domains and subject alternate names in practice, it's just that the AWS API requires to specify a primary domain. I don't think it's worth documenting this implementation detail in the variable description (a comment on the resource could be helpful though).

terragrunt/modules/acm-certificate/main.tf
```diff
@@ -0,0 +1,41 @@
+locals {
+  top_level_domains = { for domain in var.domains : domain => join(".", reverse(slice(reverse(split(".", domain)), 0, 2))) }
```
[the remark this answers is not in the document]
pietroalbini: This gets a list of the top-level domains for the domains included in the certificate (like `crates.io` and `rust-lang.org` for `static.crates.io` and `static.rust-lang.org`). That's needed to get the correct DNS zones.

terragrunt/modules/crates-io/cloudfront-webapp.tf
pietroalbini: One year :sweat_smile: Yeah we need the comment :P

terragrunt/modules/crates-io/iam.tf
pietroalbini: Not sure if IAM supports dots.

terragrunt/modules/acm-certificate/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  domain_name               = var.domains[0]
+  subject_alternative_names = slice(var.domains, 1, length(var.domains))
+  validation_method         = "DNS"
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_route53_record" "cert_validation" {
```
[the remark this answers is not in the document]
pietroalbini: I think so. The DNS record is not the record of the service, but the record used for validating the ownership of the domain when issuing the certificate. For example:

```
_2b22bc60aa422e56d75b13e0214d82c5.staging.crates.io => _2d2b066a0344d52e0467384209b47a07.mzlfeqexyx.acm-validations.aws.
```

terragrunt/modules/acm-certificate/main.tf
```diff
@@ -0,0 +1,41 @@
...
+}
+
+resource "aws_acm_certificate" "cert" {
+  domain_name               = var.domains[0]
+  subject_alternative_names = slice(var.domains, 1, length(var.domains))
+  validation_method         = "DNS"
+
+  lifecycle {
+    create_before_destroy = true
+  }
```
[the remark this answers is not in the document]
pietroalbini: The lifecycle is used when changing the certificate (which requires deleting it and creating a new one). Since the certificate is in use (by Cloudfront or a load balancer, depending on where the module is used), we can't delete it before a new one is created, we have to create a new one, switch the resources to use the new one, and then delete the old one. That's what the lifecycle does.

terragrunt/modules/crates-io/s3-static.tf
pietroalbini: That's mostly used as a backup and as a fallback when the main AWS region is down, so that we can failover to it. See https://ops-guide.crates.io/on-call/common/dl-endpoint.html

terragrunt/modules/crates-io/iam.tf
rylev: I don't mean using the domain itself but rather using the domain as the input that gets transformed into something using `_`.

PR #168, opened 2022-12-12
Use profiles and generate Terragrunt variables with a Python script

pietroalbini:
Up until now, deployment of the Terragrunt configuration was done by setting the `AWS_PROFILE` environment variable to the profile for the account. That's annoying, and as @rylev discovered it kinda breaks when you want to access multiple accounts.

The AWS Terraform provider and the Terragrunt remote state configuration both support specifying a profile, so in theory that should solve the problem if we specify the profile there. The problem is, we use `get_aws_account_id()` to create the state bucket name, and unfortunately [that function doesn't read the profile name from the remote state configuration](https://github.com/gruntwork-io/terragrunt/issues/791): it always uses the credentials for the environment, which results in the wrong bucket name.

To solve the issue, this PR uses the `run_cmd` Terragrunt function to load the variable contents from a Python script. That allows us to do whatever processing we want to get Terragrunt variables, including getting the correct account ID for our profile.

Since the script runs before any code touching AWS, in a **future** PR it could also be expanded to automatically populate your `~/.aws/config` with the appropriate profile definitions.

## Implementation notes

I originally wanted to create *another* `terragrunt.hcl` that deals with the Python script at the root of the repository, so that a state file's `terragrunt.hcl` imports an account's `terragrunt.hcl` that imports the root `terragrunt.hcl`. Unfortunately that's not possible, because Terragrunt doesn't support nesting imports.

Instead, to avoid duplicating the `remote_state` block all over the place, I replaced the per-account `terragrunt.hcl` with an `account.json` file that's read by the Python script. I use JSON there mostly because it's in the Python stdlib and doesn't require extra dependencies, but I don't like it either (suggestions welcome).

I had to take special care on the `dev-desktops-prod`, since they included a variable shared between all of their state files. I replaced that with a separate Terragrunt file that's also imported, and that generates the provider configuration rather than inputting the subscription ID.


jdno (approved):
I think this is a good starting point for us. It would definitely be great to select the right profile automatically, and if Terragrunt does not support that (yet?) the Python script is still manageable.

Generally speaking, I'm a little bit concerned adding more custom tooling around Terragrunt and Terraform, since it will add technical debt and might make it more difficult to upgrade in the future.

PR #169, opened 2022-12-13
Document terragrunt modules

jdno:
While working on the first modules in Terragrunt, we noticed a need to define a shared terminology and set some ground rules. Especially the question of composition has been discussed in depth, and the resulting decision has been documented in the section on state vs. modules.

[Rendered view](https://github.com/rust-lang/simpleinfra/blob/536916d8c0ab0d1a146f08a3ff7a218594c111d4/terragrunt/README.md)

### DIFF
--- a/terragrunt/README.md
+++ b/terragrunt/README.md
@@ -1,20 +1,41 @@
 # Terragrunt
 
-This directory contains all of the infrastructure configuration controlled through terragrunt.
+This directory contains all the infrastructure configuration controlled through
+Terragrunt.
 
-## What and why terragrunt?
+## What and why Terragrunt?
 
-[Terragrunt](https://terragrunt.gruntwork.io/) is a wrapper around terraform that adds some additional functionality that we find useful. The features of terragrunt we use are:
-* Configuration of s3 and dynamodb to handle remote terraform state files. This is done automatically instead of needing to be configured manually as is the case when using plain terraform.
-* Module reuse. Terragrunt allows for different environments (e.g., staging and production) which share almost all of their configuration. When using plain terraform this would require copy/pasting configuration and updating configuration in multiple places.
+[Terragrunt] is a wrapper around [Terraform] that adds some additional
+functionality that we find useful. The features of Terragrunt we use are:
 
+  - [Managed state files](https://terragrunt.gruntwork.io/docs/getting-started/quick-start/#keep-your-backend-configuration-dry):
+    Terragrunt can create the S3 bucket and DynamoDB tables that Terraform needs
+    to store its state. Automating their creation ensures consistency across
+    environments.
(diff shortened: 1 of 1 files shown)


pietroalbini (approved):

terragrunt/README.md
```diff
@@ -1,20 +1,122 @@
...
-## Directory structure
+## Directory Structure
 
-The `terragrunt` directory is split into two subdirectories:
-* `modules`: the reuseable module definitions for the logical services we manage.
-* `aws`: AWS account definitions which use the modules in the `modules` directory to configure the various AWS accounts we manage.
+The [Terragrunt] configuration is split into two parts: _modules_ and _state_:
 
-## Running
+- _Modules_ are normal [Terraform] modules that each provision a specific part
```
rylev, line 24: I think we might not want to say that the "provision" but rather than the provide the configuration for provisioning. Then we should mention that _state_ use these modules as to do the provisioning.

terragrunt/README.md
```diff
@@ -1,20 +1,122 @@
...
+accounts
+└── dev-desktops-prod
+    ├── westeurope
+    └── westus2
+```
+
+`dev-desktops-prod` is (AWS) _account_ that stores the Terraform state files for
+this environment. `westeurope` and `westus2` are two _states_, i.e. concrete
+instances of the [`dev-desktops-azure`](./modules/dev-desktops-azure) module.
+
```
rylev, line 54: Perhaps we should clarify why an AWS account houses two instances of Azure based modules - i.e., that the shared terraform state lives in the AWS account.

terragrunt/README.md
```diff
@@ -1,20 +1,122 @@
...
+our [dev desktops](./modules/dev-desktops-azure) that contains every piece of 
+infrastructure that is required to run one or more instances on Azure.
+
+_Services_ can be deployed multiple times. Each deployment is a concrete
+instance of the configuration, and is managed with [Terragrunt] as a _state_
+inside the [`accounts`](#accounts) directory. For example, `crates.io` has a
+staging and a production _state_, which are both configured through the
+[`crates-io`](./modules/crates-io) _service_ module.
+
+Modules that are generic and can be shared between _services_ are stored in the
```
rylev, line 72: I thought the plan was to get rid of the shared directory. Master does not have a shared directory and I changed #172 to not use one.

jdno: The reason there is no `modules/shared` directory on `master` is that I thought we were going to use dependencies in Terragrunt when I migrated `crates-io`. But my understanding from our conversation was that we want to use `modules/` for anything that can be a state in Terragrunt, and `modules/shared` for internal dependencies/resources in Terraform.

terragrunt/README.md
jdno: Added a sentence below.

rylev (approved):

PR #171, opened 2022-12-13
Create script to migrate state

jdno:
A Python script has been added that can be used to migrate state from Terraform to terragrunt. It takes the name of a Terraform module as an argument, and tries to import this module into the terragrunt configuration in the current directory. Resources are limited by a mandatory prefix, e.g. `module.staging`, to split and rename the resources.

The script uses `terragrunt import` to migrate the resources. This means no existing infrastructure or state is changed, but instead a duplicate state is built up in terragrunt. Some resources and attributes cannot get imported, though, and a `terragrunt apply` might still be required at the end.

### DIFF
--- a/terragrunt/import-state.py
+++ b/terragrunt/import-state.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python3
+
+# Import Terraform state into terragrunt
+#
+# This script can be used to import the state of a given Terraform module into
+# the terragrunt configuration in the current directory. The script only
+# performs non-destructive imports into the terragrunt state, and leaves the
+# original Terraform state untouched.
+
+import argparse
+import json
+import os
+import subprocess
+
+from collections import namedtuple
+from typing import Optional
+
+
+Config = namedtuple(
+    "Config",
+    [
+        "terraform_module_path",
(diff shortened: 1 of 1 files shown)


jdno:
This has been very helpful when migrating `crates-io` into our Terragrunt configuration. But it also has some assumptions baked in and is not that flexible. I'm honestly not sure if we want to add this to the repository or not.

jdno:
And it doesn't support submodules (e.g. `module.staging.module.certificate`). Which we'll probably need for other migrations.

jdno:
And, finally, we might want to copy the state file and then rename resources in-place to avoid running `apply` at all.

terragrunt/import-state.py
```diff
@@ -0,0 +1,296 @@
+#!/usr/bin/env python3
+
+# Import Terraform state into terragrunt
```
rylev, line 3: Can we show an example run of this command?

rylev (approved):
This looks fine to me. I think having something like this is useful even if additional steps are needed to do a full import.

pietroalbini (approved):

PR #172, opened 2022-12-13
docs-rs Staging Environment ECS Cluster

rylev:
This adds a staging environment for docs-rs. In this PR we are only adding the ECS cluster. The rest of the infra will be added in a follow up.

Some notes of interest:
* The ECS module creates its own VPC. We could move the VPC out if we believe we will be running more things than the ECS cluster in the VPC.
* There is a state file that handles the DNS Zone used by the docs-rs-staging account. All other state files that need to configure DNS records should depend on that state file

This still needs plenty of documentation work so I'm marking this as an in-progress PR for now.

*Thank you to @pietroalbini and @jdno for their help!*


jdno (commented):
Thanks, @rylev! This is a great starting point.

I ignored `modules/shared` for the most part in my review, since those are probably copied & pasted from `terraform/modules/shared`?

terragrunt/modules/dns-zone/outputs.tf
```diff
@@ -0,0 +1,9 @@
+output "name" {
```
pietroalbini, line 1: To avoid too many files, I'd move the outputs (and the inputs) into `_terraform.tf`.

terragrunt/modules/ecs-cluster/vpc.tf
```diff
@@ -0,0 +1,19 @@
...
+    1 = "usw1-az3",
+  }
+  private_subnets = {
+    2 = "usw1-az1",
+    3 = "usw1-az3",
+  }
+  untrusted_subnets = {
+    4 = "usw1-az1",
+    5 = "usw1-az3",
+  }
```
pietroalbini, line 18: For now, let's have only a public subnet and a private subnet with no untrusted subnets to save money on NAT gateways.

terragrunt/modules/dns-zone/outputs.tf
jdno: The [Standard Module Structure](https://developer.hashicorp.com/terraform/language/modules/develop/structure) for Terraform modules explicitly names `main.tf`, `variables.tf`, and `outputs.tf`. For stuff in `modules/shared`, I think it's helpful to follow that convention.

terragrunt/modules/ecs-cluster/vpc.tf
```diff
@@ -0,0 +1,19 @@
+module "vpc" {
```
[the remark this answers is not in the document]
rylev: Done

jdno (approved):
With the VPC in it's own state, this is looking good! 👏

I'm happy to have this merged so that we can get back to smaller pull requests that are easier to review. 😅

PR #173, opened 2022-12-15
Increase quota on dev-desktops

jdno:
A few users have reported running out of disk space on the new dev desktops, which has a default file system quota of 50GB per user. Increasing the quota to 100GB has resolved the issue for most users, so we want to make this the new default to reduce friction for users and work for the infra team.

With the increase in quota, two scripts have been added that allow administrators to report the current quotas (`report-quotas.sh`) and to increase them (`increase-quotas.sh`). These can be used on existing machines to adjust the quota for all users to a new value.


jdno:
I've tested the scripts manually on one of the new Azure desktops. Will roll them out to all machines and update the quotas after this PR has been approved and merged.

ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -93,6 +93,16 @@
     - "quotaon -uv /"
   when: user_quota_check.rc == 0
 
+- name: Copy the scripts to manage quotas
```
Mark-Simulacrum, line 96: Hm, I feel like I'd rather see these run by Ansible (automatically on apply), rather than needing to login ourselves to update a particular machine.

Happy to see this merged in the meantime though.

ansible/roles/dev-desktop/templates/increase-quota.sh
```diff
@@ -0,0 +1,90 @@
...
+  echo "Invalid soft limit: ${soft_limit}. Example: 50G"
+  exit 2
+fi
+
+# Get list of users
+users=$(cut -d: -f1 /etc/passwd | grep gh-)
+
+# Set quota
+for user in $users; do
+  cmd="setquota -u ${user} ${soft_limit} ${hard_limit} 0 0 /"
```
Mark-Simulacrum, line 82: How do we update the default quota for new users, created after this script runs? Is that done with the Ansible value for the filesystem quota? Can we document that at the top of this file?

ansible/roles/dev-desktop/tasks/main.yml
jdno: Oh I didn't even consider running this through Ansible. The benefit would be that we can then fully rely on a single variable to set the quota for both existing and new users. Let me give that a try...

jdno: Thanks for catching this, @Mark-Simulacrum! Running this through Ansible is actually much simpler and offers much better maintainability. 🙏

ansible/roles/dev-desktop/templates/increase-quota.sh
Mark-Simulacrum: I don't think I quite understand how we've resolved this yet FWIW

jdno: You got me thinking. Previously, we used the `team_login` binary to set the quota on new accounts. But that felt kinda dirty, so I went ahead and refactored that to instead use the `QUOTAUSER` setting in Linux. The quotas should not be fully managed by Ansible.

pietroalbini (approved):

PR #175, opened 2022-12-20
Update list of infra-admins

jdno:
aidanhs has left the infra-team, and jdn has joined the infra-admins.


rylev (commented):
Can we bring the list of admins in from somewhere else so that we don't have to repeat it in all the config files?

jdno:
I'm not sure if there was a reason that this was defined on a per-playbook level. @pietroalbini, do you know more about this?

pietroalbini:
We can definitely have the list of admins somewhere, even though we should allow to have extra sudo users (and not hardcode it to just admins). Maybe set the list in `group_vars/all.yml`?

jdno:
Done. :+1:

rylev (approved):

pietroalbini:
Deployed across all playbooks.

PR #179, opened 2022-12-22
Fix terragrunt-locals for the legacy account

pietroalbini:
Turns out the `default` profile doesn't load credentials from the environment. This PR updates the script to also support `null`, which will omit the profile.


jdno (approved):
Thanks! 👍

PR #180, opened 2022-12-22
Set up billing access for the Rust Foundation

jdno:
Some members of the Rust Foundation need access to the billing portal on the new account. This access has been granted using an AWS-managed policy.


jdno:
I deliberately didn't migrate the existing roles and policies from [`foundation.tf`](https://github.com/rust-lang/simpleinfra/blob/master/terraform/team-members-access/foundation.tf), but focused only on the billing access here.

pietroalbini (approved):

PR #182, opened 2022-12-29
cloud-dev: Add some more common dependencies

jyn514:
Most of these are self-explanatory. Powershell is useful for testing changes to `./x.ps1`. `delta` is a program for making `git diff` easier to read.

### DIFF
--- a/ansible/roles/dev-desktop/tasks/main.yml
+++ b/ansible/roles/dev-desktop/tasks/main.yml
@@ -197,7 +197,7 @@
   notify:
     - restart-firewall
 
-- name: install common tooling
+- name: Install common tooling
   apt:
     name:
       - build-essential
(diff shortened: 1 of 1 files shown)


ansible/roles/dev-desktop/tasks/main.yml
```diff
@@ -209,8 +218,18 @@
...
+      - jq
+      - powershell
     state: present
 
+- name: Install git-delta
+  shell: |
+    DELTA_VERSION=$(curl --silent https://api.github.com/repos/dandavison/delta/releases/latest | jq --raw-output '.name')
+    curl -O git-delta.deb https://github.com/dandavison/delta/releases/download/$DELTA_VERSION/git-delta_$DELTA_VERSION_$(uname -m).deb
+    apt install ./git-delta.deb
+    rm ./git-delta.deb
```
jyn514, line 231: admittedly this is kinda janky; I'm ok with installing it just for my user if you don't think other people on the machine will use it.

jdno (commented):
I suggest that we split the Powershell and git-delta into their own PRs, and get this merged to deploy the other tools asap.

ansible/roles/dev-desktop/tasks/main.yml
jyn514: 👍 makes sense, I don't think a lot of other people are using it, happy to remove it

jyn514:
Ok, pushed a change removing powershell and `git-delta` :)

jdno (approved):

PR #183, opened 2022-12-29
cloud-dev: don't assume scripts are always run from the home directory

jyn514:


jdno:
I think the assumption was that scripts are always run in the current directory. Theoretically (read: I haven't tested this myself yet), it should be possible to set up different directories with different clones and worktrees. Did this not work for you?

jyn514:
@jdno my use case was to be able to run this from inside an existing checkout, which is how I normally add a worktree:
```
(bash@dev-desktop-us-1.infra.rust-lang.org) ~/rust0 [16:40:28]
> git worktree add ../rust-example
Preparing worktree (new branch 'rust-example')
Updating files: 100% (39002/39002), done.
HEAD is now at 23d3ee81867 Revert "bootstrap: Get rid of `tail_args` in `stream_cargo`"
```
With these scripts, that gives an error:
```
> new_worktree.sh 
++ ls
++ grep -E -e 'rust[0-9]+'
++ wc -l
+ N=0
+ echo 0
0
+ pushd rust
/usr/local/bin/new_worktree.sh: line 7: pushd: rust: No such file or directory
```

More than that though, I think they *do* assume the home directory: this line in particular assumes you have a `config.toml` in the parent directory, which is only present in the home directory from `/etc/skel`, and it's not documented anywhere that you should create it in a subdirectory before running the script: https://github.com/rust-lang/simpleinfra/blob/68705397231b1d3309b167cbc728cd0ec6782323/ansible/roles/dev-desktop/files/scripts/new_worktree.sh#L14

If you want, I can instead change them to see if they're being run from an existing `rust` checkout and cd to the parent directory instead?

jdno:
Sorry for the long silence, @jyn514. 🙏 

I've looked through the scripts, and `set_defaults.sh` does the following:
https://github.com/rust-lang/simpleinfra/blob/cb143cd0b1dba3099ab5013898b8e241aee371fd/ansible/roles/dev-desktop/files/scripts/set_defaults.sh#L12

If we do the same in `new_worktree.sh`, I _think_ that we should be able to run the scripts from anywhere in `~/`. Would that make sense?

jyn514:
I'm not sure how config.toml is related? the problems here are because the scripts assume they are being run in the *parent* directory of all the rust checkouts, which is not the case, people can run them from anywhere.

jdno:
🤦‍♂️ My bad. I got stuck on the issue with the home directory and read over the problem with the parent directory. I think you're assessment that the scripts assume to be run from `~/` is correct.

Hard-coding this assumption into the scripts is fine for me for now, and we can revisit this if it turns out to be a problem for someone else's workflow. In that case it might make more sense to check whether you're in a checkout or not, but let's cross that bridge when we get there.

jdno (approved):

PR #184, opened 2023-01-04
Deploy Fastly function for static.crates.io

jdno:
The Fastly function for static.crates.io has been deployed. No traffic is routed to Fastly yet, but this allows us to start testing internally.


rylev (approved):
It might make sense in the PR description to link to the commit being deployed for easier review.

jdno:
That's a good idea. In this case, I set it to the [latest commit](https://github.com/rust-lang/simpleinfra/commit/0ab29ba22374a2ceafdce4540b45a5030f6cc9c2) on `master` since the actual changes were so long ago.

PR #185, opened 2023-01-04
Add remaining docs-rs-staging infra except for ecs app

rylev:
This adds the remaining docs-rs-staging infra with the exception of the ecs app which I would like to do in a follow-up PR. 

Since we do not have an ecs app configured the cloud front distribution is pointed to example.com as its origin. Once the ecs app is configured, we will change the origin to the ecs app. 

Also, the `aws_s3_bucket_inventory` used for inventory storage of the main s3 bucket is now configurable through a variable. This variable determines whether to do the inventory storage or not. As is currently configured, docs-rs-staging will *not* have a perform this inventory storage.


terragrunt/modules/docs-rs/s3.tf
```diff
@@ -0,0 +1,106 @@
+resource "aws_s3_bucket" "storage" {
+  bucket = "static-docs-rs"
```
rylev, line 2: Needs to be globally unique. Perhaps post fix with the account id.

terragrunt/modules/docs-rs/s3.tf
```diff
@@ -0,0 +1,106 @@
...
+
+  included_object_versions = "Current"
+  optional_fields          = ["ETag", "ReplicationStatus", "Size"]
+
+  schedule {
+    frequency = "Weekly"
+  }
+  destination {
+    bucket {
+      bucket_arn = var.inventories_bucket_arn
```
rylev, line 47: Just create the bucket. 

Question: does this need to be configurable?

terragrunt/accounts/docs-rs-staging/docs-rs/terragrunt.hcl
```diff
@@ -0,0 +1,23 @@
+terraform {
```
rylev, line 1: Ask in Zulip whether we need different states for the web and the builder (considering they share resources like s3 and RDS).

terragrunt/modules/docs-rs/s3.tf
```diff
@@ -0,0 +1,106 @@
...
+  bucket = aws_s3_bucket.storage.id
+
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
+resource "aws_s3_bucket" "backups" {
+  bucket = "docs.rs-backups"
```
rylev, line 64: Make name globally unique

terragrunt/modules/docs-rs/s3.tf
```diff
@@ -0,0 +1,106 @@
+resource "aws_s3_bucket" "storage" {
```
rylev, line 1: Change this name to "static-storage"

terragrunt/accounts/docs-rs-staging/docs-rs/terragrunt.hcl
rylev: It was agreed in Zulip that this does not need to happen. We should keep them in the same state for simplicity sake.

jdno (approved):
Thanks, @ryan!

rylev:
@jdno I had to push a new commit to get things work. This is now all applied to the docs-rs-staging account. Please have a look and let me know if I need to adjust in a new PR.

PR #190, opened 2023-01-06
Remove packages missing in apt

jdno:
A few of the new packages for the dev-desktops are not distributed through apt. They have been removed for now, and will be added back in a later commit after some refactoring.


jdno:
cc @jyn514

PR #191, opened 2023-01-06
Make acm-certificate a sharable Terraform module

jdno:
The acm-certificate module was falsely created as a service in Terragrunt. But it is better suited to be a Terraform module that other modules can include.

Due to slight differences between the legacy account and new accounts, a new variable is necessary that changes how the DNS zone for the certificate is calculated.


pietroalbini (approved):

terragrunt/modules/acm-certificate/variables.tf
```diff
@@ -2,3 +2,8 @@ variable "domains" {
   description = "List of domain names included in the certificate"
   type        = list(string)
 }
+
+variable "legacy" {
```
rylev, line 6: I'm not sure this is the right variable. It might be more straight-forward to make the variable more directly about what it's used for.

jdno: My reasoning was that should only be set to `true` in the legacy account. 😅

rylev: I get it - makes the code harder to understand IMO, but it's also not a big deal.

jdno: I'm happy to improve this still. Just wanted to get this merged first to unblock more pull requests. 😅 Although I have to admit that I'm not sure how this can be made significantly better. More documentation maybe?

PR #198, opened 2023-01-11
docs-rs staging app

rylev:
Adds the ecs app which runs the docs-rs service


rylev:
Going to merge this. I've already did a light review with @jdno, and we didn't find any major issues. Also, this is mostly a copy/paste job from the terraform directory with small changes. This is also deployed already. It's probably easier to adjust in follow-up PRs rather than keep this open.

terragrunt/modules/docs-rs/web-server.tf
```diff
@@ -1,88 +1,92 @@
...
-#   source         = "../shared/modules/ecs-app"
-#   cluster_config = data.terraform_remote_state.shared.outputs.ecs_cluster_config
+module "web" {
+  source         = "../ecs-app"
+  cluster_config = var.cluster_config
 
-#   env  = "prod"
-#   name = "docs-rs-web"
-#   repo = "rust-lang/docs.rs"
+  env  = "dev"
```
jdno, line 14: Why are we not using `staging` here to keep naming consistent with the account?

terragrunt/modules/docs-rs/web-server.tf
```diff
@@ -1,88 +1,92 @@
...
+    DOCSRS_LOG                  = "docs_rs=debug,rustwide=info"
+    RUST_BACKTRACE              = "1"
+    DOCSRS_STATIC_CLOUDFRONT_ID = "${aws_cloudfront_distribution.static.id}"
+  }
 
-#   secrets = {
-#     DOCSRS_GITHUB_ACCESSTOKEN = "/prod/docs-rs/github-access-token"
-#   }
+  secrets = {
+    DOCSRS_GITHUB_ACCESSTOKEN = "/prod/docs-rs/github-access-token"
```
jdno, line 33: I think that we should have a `/staging` namespace now that we're spinning up staging environments. Should we create a GitHub issues for this?

terragrunt/modules/docs-rs/web-server.tf
```diff
@@ -1,88 +1,92 @@
...
+          "s3:PutObject",
+          "s3:GetObjectTagging",
+          "s3:PutObjectTagging",
+          "s3:GetObject",
+          "s3:PutObjectAcl",
+          "s3:ListBucket",
+          "s3:DeleteObject",
+        ]
+        Resource = [
+          "arn:aws:s3:::rust-docs-rs",
```
jdno, line 74: Are we sharing the S3 bucket between staging and production? This feels like something that should come from either a variable or a resource.

terragrunt/modules/ecs-app/ci.tf
```diff
@@ -0,0 +1,44 @@
...
+
+// IAM User used by GitHub Actions to pull and push images to ECR, and to
+// restart the ECS service once the image is uploaded. The credentials will
+// be added automatically to the GitHub Actions secrets.
+
+module "iam_ci" {
+  source     = "../gha-iam-user"
+  org        = split("/", var.repo)[0]
+  repo       = split("/", var.repo)[1]
+  env_prefix = var.env == "dev" ? "DEV" : null
```
jdno, line 16: Same question here as above regarding the naming of the environment: `dev` vs. `staging.`

terragrunt/modules/ecs-app/deployment.tf
```diff
@@ -0,0 +1,227 @@
...
+    {
+      name      = "app"
+      image     = module.ecr.url
+      essential = true
+
+      logConfiguration = {
+        logDriver = "awslogs"
+        options = {
+          "awslogs-group"         = aws_cloudwatch_log_group.task.name
+          "awslogs-region"        = "us-west-1"
```
jdno, line 130: Should we change this to `us-east-2` based on our [decision re. AWS regions](https://forge.rust-lang.org/infra/docs/aws-regions.html)?

jdno (commented):
I've added a few comments that would be great to address, but that I wouldn't consider blocking. If you decide to go ahead and merge this, can you either create GitHub issues for the open comments or ping me so that I can do it?

terragrunt/modules/docs-rs/web-server.tf
jdno: This might sound more passive-aggressive than it was intended. 😅 I suspect there might be a historic reason why `dev` is used here. But if there isn't, I would strongly advise us to keep the naming consistent between environment variables and AWS accounts.

rylev: That's a good point. `staging` is a much longer word, but perhaps that's fine?

terragrunt/modules/docs-rs/web-server.tf
rylev: This should certainly not be in a prod namespace. However, I think we're not going to have multiple environments in the same account so perhaps we should just drop this completely?

terragrunt/modules/docs-rs/web-server.tf
rylev: I believe these will be in separate accounts.

rylev:
@jdno I've addressed all of your comments and they're deployed. Going to merge this.

terragrunt/modules/docs-rs/web-server.tf
jdno: Oh that's a good point! :+1:

terragrunt/modules/docs-rs/web-server.tf
jdno: Bucket names are globally unique. So it doesn't matter in what account this is, it would still refer to the one `rust-docs-rs` bucket. Not the first time we're running into this. 😞

PR #199, opened 2023-01-11
Updates to AWS MFA policies due to changes in AWS implementation and best practices

carols10cents:
## What happened

I tried to change my MFA in my AWS account. When I tried, I got this error:

```
User: arn:aws:iam::890664054962:user/carols10cents is not authorized to perform: iam:CreateVirtualMFADevice on resource: arn:aws:iam::890664054962:mfa/aaa because no identity-based policy allows the iam:CreateVirtualMFADevice action
```

despite [appearing to have that permission](https://github.com/rust-lang/simpleinfra/blob/f55b3b921d09123d00aaa7dfb4356f5efe75dfc6/terraform/team-members-access/_shared.tf#L80). 

I also tried logging out and logging in to no avail.

After searching, I found [this AWS question and answer from 2 months ago](https://repost.aws/questions/QUBD8nxw68S92EaaPiCydisg/how-to-allow-mfa-self-management-with-custom-mfa-device-names-multiple-mfa), which indicates AWS only recently began allowing users to have more than one MFA device and thus give them custom names.

To test that theory, I added an MFA device named `carols10cents` and it worked.

So the first commit in this branch removes the restriction that the MFA device be named with your username 🤦🏻‍♀️ 

The second commit is a change I noticed when reading [the updated documentation for MFA IAM permissions](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws_my-sec-creds-self-manage-mfa-only.html) in a big "warning" box. I can't claim to understand this completely, but AWS is recommending this, so... 🤷🏻‍♀️ 

Please review this carefully because I only barely know what I'm doing here!!! I'm happy to make whatever changes.

### DIFF
--- a/terraform/team-members-access/_shared.tf
+++ b/terraform/team-members-access/_shared.tf
@@ -80,7 +80,7 @@ resource "aws_iam_policy" "manage_own_credentials" {
         "iam:CreateVirtualMFADevice",
         "iam:DeleteVirtualMFADevice"
       ],
-      "Resource": "arn:aws:iam::*:mfa/$${aws:username}"
+      "Resource": "arn:aws:iam::*:mfa/*"
     },
     {
       "Sid": "AllowManageOwnUserMFA",
--- a/terraform/team-members-access/_shared.tf
+++ b/terraform/team-members-access/_shared.tf
@@ -77,8 +77,7 @@ resource "aws_iam_policy" "manage_own_credentials" {
       "Sid": "AllowManageOwnVirtualMFADevice",
       "Effect": "Allow",
       "Action": [
-        "iam:CreateVirtualMFADevice",
-        "iam:DeleteVirtualMFADevice"
+        "iam:CreateVirtualMFADevice"
       ],
       "Resource": "arn:aws:iam::*:mfa/*"
     },


jdno:
Hi @carols10cents,

Thanks for digging up all of this information, and sorry that you had to resolve your issue yourself. 😭 

I've started looking into this and will run a few experiments tomorrow. From my initial impression, we _should_ be able to keep `DeleteVirtualMFADevice` if we add the `Condition` in one of the linked sites. The warning seems to refer to the `DenyAllExceptListedIfNoMFA` list, which doesn't include this permission..

Mark-Simulacrum:
I think as part of this we should ask whether it's *right* that you can delete + re-create MFA without having a current MFA device, it seems -- that seems like basically equivalent to not having MFA enforced. So probably ought to be changed to require an admin to re-enable you or something.

carols10cents:
When I have gotten a new phone in the past, I've had to disable MFA on accounts and then re-enable it on my new phone. I don't think I'll personally have to do that again because I'm now using BitWarden that can manage MFA independently of device, but that is an issue.

Do the AWS admins want to have to deal with these potential problems?

jdno:
> I think as part of this we should ask whether it's right that you can delete + re-create MFA without having a current MFA device

As far as I understand the current implementation, without an MFA device you can only login and create a new one. Deleting an MFA device actually requires you to be signed in with one (since it's not in the `DenyAllExceptListedIfNoMFA` list). So afaik leaving that permission in place would be fine from a security perspective.

> Do the AWS admins want to have to deal with these potential problems?

I expect these to happen rarely enough that it would be fine to require admins to re-create MFA devices for users.

So in summary, I'm fine with both approaches. I think we can set up the permissions so that you always need an MFA device, but I'm equally happy to add an escalation to admins for anything that involves changing MFA.

Mark-Simulacrum:
> without an MFA device you can only login and create a new one

Right, but if AWS allows more than one (as is now the case), then "escalating" to regular access seems like it can be done without current MFA? If we condition the ability to add MFA on not having any MFA configured, then that seems more workable.

> I expect these to happen rarely enough that it would be fine to require admins to re-create MFA devices for users.

:+1: -- we should not be changing MFA that often, I think.

pietroalbini:
I agree we should get infra involvement to re-create MFA.

jdno (approved):
> then "escalating" to regular access seems like it can be done without current MFA

Oh that is a great point that I didn't consider. I'm totally on board now, let's go through infra if MFA needs to be changed. I think that means we can just merge this pull request.

I made to note to check Forge and update the documentation if necessary. I'll take a look at that sometime this week.

PR #200, opened 2023-01-12
Add configuration for a bastion server in docs-rs-staging

rylev:
This allows users with the right permissions to access things on the private subnets (e.g., the RDS instance).

### DIFF
--- a/terragrunt/modules/bastion/instance.tf
+++ b/terragrunt/modules/bastion/instance.tf
@@ -35,7 +35,7 @@ resource "aws_route53_record" "bastion" {
 
 // Create the EC2 instance itself.
 
-data "aws_ami" "ubuntu_bionic" {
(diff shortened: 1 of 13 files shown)


terragrunt/modules/bastion/instance.tf
```diff
@@ -0,0 +1,79 @@
...
+  zone_id = var.zone_id
+  name    = "bastion.${data.aws_route53_zone.zone.name}"
+  type    = "A"
+  records = [aws_eip.bastion.public_ip]
+  ttl     = 300
+}
+
+// Create the EC2 instance itself.
+
+data "aws_ami" "ubuntu_bionic" {
```
Mark-Simulacrum, line 38: Bionic seems off here - we're using focal below. Probably worth moving to 22.04 as well, but let's at least name the resource right (maybe just "Ubuntu").

PR #203, opened 2023-01-17
Create weighted records for static.crates.io

jdno:
Weighted records[^1] allow us to manage traffic between different targets. We use them for static.crates.io to split traffic between CloudFront and Fastly. The exact distribution can be set via variables for each environment.

[^1]: https://docs.aws.amazon.com/Route53/latest/APIReference/api-policies-traffic-policy-document-format.html#traffic-policy-document-format-rules-weighted

### DIFF
--- a/terragrunt/accounts/legacy/crates-io-staging/crates-io/terragrunt.hcl
+++ b/terragrunt/accounts/legacy/crates-io-staging/crates-io/terragrunt.hcl
@@ -22,4 +22,7 @@ inputs = {
   iam_prefix = "staging-crates-io"
 
   strict_security_headers = true
+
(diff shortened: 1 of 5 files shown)


jdno:
The new alias in the certificate sadly requires the re-creation of the certificate. And there can be a slight timing issue with the DNS records, where the weighted records cannot be created because Route53 still thinks a record for `static.crates.io` exists without a weight. I've tried fixing this by specifying dependencies between the `aws_route53_record.cloudfront_static_domain` and `aws_route53_record.weighted_static_cloudfront` records, but that didn't solve the issue.

There might thus be a small downtime when deploying this to production.

pietroalbini (approved):

PR #207, opened 2023-01-18
ECS Task Module

rylev:
We are going to want to define more modules that are not always going to fit into the same shape as an HTTP based app. This allows for task definitions to be created separate from the HTTP app.


terragrunt/modules/ecs-task/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  memory       = var.memory
+  network_mode = "awsvpc"
+
+  task_role_arn            = aws_iam_role.task.arn
+  execution_role_arn       = aws_iam_role.task_execution.arn
+  requires_compatibilities = ["FARGATE"]
+
+  container_definitions = jsonencode([
+    {
+      name      = "app"
```
jdno, line 15: Do we want to set this dynamically or through a variable??

terragrunt/modules/ecs-task/main.tf
```diff
@@ -0,0 +1,74 @@
...
+      image     = module.ecr.url
+      essential = true
+
+      logConfiguration = {
+        logDriver = "awslogs"
+        options = {
+          "awslogs-group"         = aws_cloudwatch_log_group.task.name
+          "awslogs-region"        = data.aws_region.current.name
+          "awslogs-stream-prefix" = "app"
```
jdno, line 24: Same question as above.

jdno (approved):
Nice work!

rylev:
CI errors are caused by the new clippy release and is not related to changes in this PR. Fixing clippy errors in #215 and merging this.

PR #208, opened 2023-01-19
Add script for running tasks

rylev:
This is a fairly hacky script for running one-off tasks on ECS which has been very useful for testing and bootstrapping docs-rs-staging. This probably should be a python script, but I don't know Python at all so I wrote bash (which I'm terrible at). Enjoy!

### DIFF
--- a/run-task.sh
+++ b/run-task.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# This script runs an ECS task.
+# The arguments for the script are:
+# 1. The cluster name
+# 2. The task name
+# 3. The name of the container
+# 4. The names of the security groups to use as a comma separated list (with no spaces)
+# 5. The command to run
+#
+# For example:
+# 
+# 	$ ./run-task.sh docs-rs-staging docs-rs-web app docs-rs-staging-service,docs-rs-web '/usr/local/bin/cratesfyi database migrate'
+
+clusterArn=$(aws ecs describe-clusters --clusters $1 | jq -r '.clusters[0].clusterArn')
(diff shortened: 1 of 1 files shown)


jdno:
Sorry for the long delay here, Ryan! I've been giving this a lot of thought without coming to any conclusion. 😅

I feel somewhat conflicted about this. For one, we should standardize how we write scripts and utilities for ourselves. Right now, we have some Python, Rust, and Bash in the mix. I'm worried that Rust might be the most interesting language for participation and maintainability, but a bit overkill for the kind of scripts we tend to write here. This requires a broader discussion beyond the scopes of this PR, though.

The other question that's been going through my mind is what problem we're trying to solve here. Is the script something that we explicitly label as "only for the infra-team" or do we need to make the same functionality available to teams as well? Because in the latter case, I think that we should try to bring this functionality into the tools that the docs.rs team is already using. Which raises more interesting questions like "how can we maintain these integrations outside rust-lang/simpleinfra".

I don't have a good answer for either of these questions yet...

rylev:
> The other question that's been going through my mind is what problem we're trying to solve here.

This was a useful utility for me. I don't think we want folks outside of the infra team relying on these and to be honest, I don't think we want to rely on them too much ourselves. The utility of this script is short lived but it is useful in small scenarios. We can not have support for little utility scripts in simpleinfra, but then we risk siloing knowledge or re-implementing a bunch stuff more than once.

jdno (approved):
Let's merge this script for now so that we can use it for quick iteration. As we learn more about our new setup, we can see how we want to evolve this script...

PR #211, opened 2023-01-24
Set AWS region for static crates certificate

jdno:
The certificates for crates.io are managed in the us-east-1 region, while the infrastructure is deployed in us-west-2. This causes Terraform to not find the certificate, thinking that it has been removed. Setting the provider for the certificate manually resolves the issue.

### DIFF
--- a/terragrunt/modules/acm-certificate/_terraform.tf
+++ b/terragrunt/modules/acm-certificate/_terraform.tf
@@ -1,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.20"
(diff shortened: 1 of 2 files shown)


jdno:
I don't think we can do this, based on the [requirements for CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cnames-and-https-requirements.html#https-requirements-aws-region):

> To use a certificate in AWS Certificate Manager (ACM) to require HTTPS between viewers and CloudFront, make sure you request (or import) the certificate in the US East (N. Virginia) Region (`us-east-1`).

pietroalbini (approved):

PR #212, opened 2023-01-24
Enable request logging for crates.io

jdno:
As with the Rust releases (see https://github.com/rust-lang/simpleinfra/commit/8c6cf1fa9aaea8dc596103e9614c341ef942c25e), we are now collecting request logs for crates.io.

### DIFF
--- a/terragrunt/modules/crates-io/_terraform.tf
+++ b/terragrunt/modules/crates-io/_terraform.tf
@@ -91,3 +91,14 @@ variable "static_fastly_weight" {
   description = "Weight of the traffic for static.crates.io that is routed through Fastly"
   type        = number
 }
+
+variable "fastly_customer_id_ssm_parameter" {
+  description = "Name of the SSM parameter with our Fastly customer id"
+  type        = string
+}
+
+variable "fastly_aws_account_id" {
+  # See https://docs.fastly.com/en/guides/creating-an-aws-iam-role-for-fastly-logging
+  description = "The AWS account ID that Fastly uses to write logs"
+  default     = "717331877981"
+}
--- a/terragrunt/modules/crates-io/compute-static/src/log_line.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/log_line.rs
@@ -2,7 +2,7 @@ use std::net::IpAddr;
 
 use derive_builder::Builder;
 use serde::Serialize;
-use time::{Date, Time};
+use time::OffsetDateTime;
 
 #[derive(Debug, Serialize)]
 #[serde(tag = "version")]
@@ -13,8 +13,8 @@ pub enum LogLine {
 
 #[derive(Debug, Builder, Serialize)]
 pub struct LogLineV1 {
-    date: Date,
-    time: Time,
+    #[serde(with = "time::serde::rfc3339")]
+    date_time: OffsetDateTime,
     url: String,
     bytes: Option<usize>,
     ip: Option<IpAddr>,
--- a/terragrunt/modules/crates-io/compute-static/src/log_line.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/log_line.rs
@@ -1,0 +1,23 @@
+use std::net::IpAddr;
+
+use derive_builder::Builder;
+use serde::Serialize;
+use time::{Date, Time};
+
+#[derive(Debug, Serialize)]
+#[serde(tag = "version")]
+pub enum LogLine {
+    #[serde(rename = "1")]
+    V1(LogLineV1),
+}
+
+#[derive(Debug, Builder, Serialize)]
+pub struct LogLineV1 {
+    date: Date,
+    time: Time,
+    url: String,
+    bytes: Option<usize>,
+    ip: Option<IpAddr>,
+    method: Option<String>,
+    status: Option<u16>,
+}
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -43,11 +43,8 @@ fn init_logging(config: &Config) {
 
 /// Collect data for the logs from the request
 fn collect_request(request: &Request) -> LogLineV1Builder {
-    let datetime = OffsetDateTime::now_utc();
-
     LogLineV1Builder::default()
-        .date(datetime.date())
-        .time(datetime.time())
+        .date_time(OffsetDateTime::now_utc())
         .url(request.get_url_str().into())
         .ip(request.get_client_ip_addr())
         .method(Some(request.get_method().to_string()))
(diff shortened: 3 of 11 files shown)


jdno:
I've created this as a draft pull request so that we can first rollout Fastly to production. I don't want to merge new features into `master` until the rollout is live.

terragrunt/modules/crates-io/_terraform.tf
```diff
@@ -91,3 +91,14 @@ variable "static_fastly_weight" {
   description = "Weight of the traffic for static.crates.io that is routed through Fastly"
   type        = number
 }
+
+variable "fastly_customer_id" {
+  description = "Our Customer ID on Fastly"
```
jdno, line 96: Is this something that we want to add as a plain text variable or put into a secret?

terragrunt/modules/crates-io/cloudfront-webapp.tf
```diff
@@ -94,6 +94,12 @@ resource "aws_cloudfront_distribution" "webapp" {
     }
   }
 
+  logging_config {
+    include_cookies = false
+    bucket          = aws_s3_bucket.logs.bucket_regional_domain_name
+    prefix          = "cloudfront/${var.webapp_domain_name}"
```
jdno, line 100: Do we want to log all requests or just `static.crates.io`? AFAIK requests to the app should be logged already.

Mark-Simulacrum: The application itself already has request logs in Papertrail, those are probably fine for that purpose. I think just logging static.crates.io will be enough for now.

(crates.io itself serves comparatively little bandwidth, though most static.crates.io requests are redirects from crates.io I believe).

Mark-Simulacrum: (But it might make this more complicated to limit the logging -- logging all requests seems also fine to me).

terragrunt/modules/crates-io/_terraform.tf
Mark-Simulacrum: I *think* we should probably treat it as a secret; any requests coming out of the Fastly AWS account in possession of this ID can assume the role in our account. No one other than Fastly should be issuing such requests of course but no need to expose it IMO.

jdno: Yeah I feel better if we keep this private as well 👍

terragrunt/modules/crates-io/cloudfront-webapp.tf
jdno: I've removed logging from the other two CloudFront distributions and left it untouched for `static.*`.

terragrunt/modules/crates-io/compute-static/src/log_line.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Debug, Serialize)]
+#[serde(tag = "version")]
+pub enum LogLine {
+    #[serde(rename = "1")]
+    V1(LogLineV1),
+}
+
+#[derive(Debug, Builder, Serialize)]
+pub struct LogLineV1 {
+    date: Date,
```
jdno, line 16: `date` and `time` are serialized using the `serde-human-readable` feature in the `time` crate. Not sure yet if I really like that. It avoids the need for custom serialization, but it's also a bit implicit.

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -67,3 +117,31 @@ fn send_request_to_s3(config: &Config, request: &Request) -> Result<Response, Er
...
 }
+
+/// Collect data for the logs from the response
+fn collect_response(
+    log_line: &mut LogLineV1Builder,
+    response: &Result<Response, Error>,
+) -> LogLineV1Builder {
+    if let Ok(response) = response {
+        log_line
+            .bytes(response.get_content_length())
```
jdno, line 128: Do we want to rename the field to `content-length`? The reason I ask is that `HEAD` requests are still logged with the size of the content, even though no body is returned. For example:

```json
{"bytes":55252,"date":"2023-02-08","ip":"127.0.0.1","method":"HEAD","status":200,"time":"15:33:05.395478208","url":"https://fastly-static.staging.crates.io/crates/rand/rand-0.3.14.crate","version":"1"}
```

jdno:
Seems that the IAM permissions are not yet correct, since I don't see any logs in S3. I'll give that a bit of time to stream 🤞 and check it again tomorrow.

terragrunt/modules/crates-io/compute-static/src/log_line.rs
Mark-Simulacrum: Seems OK for now, but human readable sounds iffy to me. For logs I would expect us to want something less human and more machine readable...

It's not clear to me what this actually ends up being but I think something with ~no ambiguity is better.

terragrunt/modules/crates-io/fastly-static.tf
```diff
@@ -62,6 +65,24 @@ resource "fastly_service_compute" "static" {
     filename         = data.external.package.result.path
     source_code_hash = filesha512(data.external.package.result.path)
   }
+
+  logging_s3 {
+    name        = local.request_logs_endpoint
+    bucket_name = aws_s3_bucket.logs.bucket
+
+    s3_iam_role = aws_iam_role.fastly_assume_role.arn
+    domain      = "s3.us-west-1.amazonaws.com"
```
Mark-Simulacrum, line 74: I would expect us to be able to get this from the bucket somehow, but seems fine to hardcode for now.

Mark-Simulacrum: I think S3 prefers that we use bucket-in-DNS for requests though, rather than this endpoint -- e.g., fastly-requests.s3.us-west-1.amazonaws.com.

terragrunt/modules/crates-io/compute-static/src/log_line.rs
jdno: The default format is the internal integer representation of the time or date, which is impossible to interpret. I've updated the code to write dates using [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339), which should solve for both human- and machine-readability.

```json
{
  "date_time":"2023-02-28T11:17:19.002782375Z"
}
```

Mark-Simulacrum: Sounds great, thanks.

Mark-Simulacrum (approved):

terragrunt/modules/crates-io/compute-static/src/main.rs
jdno: I've been looking into this, but the `Response` struct is kinda limited. The body is streamed through the edge worker, so inspecting the size is non-trivial.

We could add a special case for `HEAD` requests like below, but we can do that when parsing the logs as well.

```rust
if request.get_method() == Method::HEAD {
    log_line.bytes(Some(0))
}
```

jdno:
I had to rebase the branch to resolve the conflicts with the recent [database dump workaround](https://github.com/rust-lang/simpleinfra/pull/250).

PR #219, opened 2023-02-01
Cleanup release scripts

kpcyrd:
I learned about this repo today and started looking into how this works. :smiley_cat: 

My vim setup highlighted some issues in the scripts that are fixed by this patch. Specifically the shell script doesn't work if your `$PWD` contains a folder with a space.

### DIFF
--- a/release-scripts/promote-release.py
+++ b/release-scripts/promote-release.py
@@ -2,10 +2,7 @@
 
 import argparse
 import json
-import os
 import subprocess
-import sys
-import tempfile
 import time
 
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,7 +61,7 @@ and you need to have `jq` and [the 1password CLI][1password-cli] installed.
 For example, to create a git tag for the Rust 2.0.0 release you can use:
 
 ```
-./with-rust-key.sh gpg tag -u FA1BE5FE 2.0.0 stable
+./with-rust-key.sh gpg tag -u 108F66205EAEB0AAA8DD5E1C85AB96E6FA1BE5FE 2.0.0 stable
```
pietroalbini, line 64: The changes in the rest of the files are fine, but I'd like to keep the shorter ID here. There is no chance of a collision happening (the `gpg` command is run inside an ephemeral `GPGHOME` with just that key in it), and the shorter ID doesn't hide the command we're trying to run.

kpcyrd: I've reverted the change in the readme :+1:

kpcyrd:
Anything I can do here? :)

pietroalbini (approved):
Sorry for the delay in getting this merged!

PR #221, opened 2023-02-02
Invite legacy account

jdno:
The legacy AWS account has been invited into the new AWS organization. This enables us to access the account using SSO, and consolidates billing for all AWS accounts.

### DIFF
--- a/terragrunt/README.md
+++ b/terragrunt/README.md
@@ -115,8 +115,37 @@ instance of the service will create its own certificate.
 
 ## Running Terragrunt
 
+When running [Terragrunt], it fetches and stores the state of the infrastructure
+in AWS. This requires that an AWS profile is configured for each account. After
+that has been done, running [Terragrunt] is as simple as replacing `terraform`
+with `terragrunt` in the command.
+
+### Configure AWS Profiles
+
 Running [Terragrunt] requires permissions to the AWS account you are
-configuring. Assuming you have permission, you can `cd` into the corresponding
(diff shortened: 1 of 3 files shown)


terragrunt/modules/aws-organization/accounts.tf
```diff
@@ -7,6 +7,11 @@ resource "aws_organizations_account" "admin" {
   email = "admin+root@rust-lang.org"
 }
 
+resource "aws_organizations_account" "legacy" {
+  name  = "Rust Admin - 8450"
```
rylev, line 11: I guess renaming is not an easy operation? This name is pretty opaque..

rylev (approved):

terragrunt/README.md
```diff
@@ -115,8 +115,37 @@ instance of the service will create its own certificate.
...
+```text
+[profile rust-root]
+sso_start_url = https://rust-lang.awsapps.com/start
+sso_account_id = <rust-root-account-id>
+sso_role_name = AdministratorAccess
+sso_region = us-east-1
+region = us-east-1
+```
+
+Each account in `./accounts` has a `account.json` file that contains the name of
```
rylev, line 138: This is only true of the root account I believe. Should we add account.json files for the other accounts?

jdno: We've added that file to all accounts in https://github.com/rust-lang/simpleinfra/pull/168.

PR #229, opened 2023-02-07
Document Terraform module for crates-io

jdno:
A README has been added to the Terraform module for `crates.io` to provide a high-level overview of the module and its resources.


rylev (approved):
I think this is a great start!

PR #235, opened 2023-02-10
Fix path error

jinmiaoluo:


jdno (changes requested):
Hi @jinmiaoluo, thanks for your contribution! 🥳

Instead of changing the link, I suggest that we'll move the README into the `terraform` directory. It looks like it got moved into `terraform/shared` when the module was refactored a few months ago.

Can you move the file and then update the link to it in the repository's [README](https://github.com/rust-lang/simpleinfra/blob/master/README.md)?

jinmiaoluo:
Hi, @jdno 

You are right. I have moved [README](https://github.com/rust-lang/simpleinfra/blob/b09ca89828f9c0c7b73534ebfc593555e0b8f3f5/terraform/shared/README.md) from `terraform/shared` to `terraform`, and reverse the change in README of dns module.

jdno (approved):
Awesome, thanks a lot! 🎉

PR #242, opened 2023-02-14
Add initial support for packer

rylev:
This adds initial support for running packer with our Ansible setup. 

Since the way we handle Ansible configuration is slightly unorthodox (i.e., we have the apply script which allows us to handle multiple environments more easily), we have created a wrapper for packer that sets up the Ansible environment properly. Running this wrapper looks like this:

```bash
$ ./packer staging docs-rs-builder --var "revision=SOME_REVISION5"
```

The script works by creating a python virtual environment and copying the proper Ansible configuration into a directory (much in the same way that the `apply` script does - in fact, I modified the `apply` script to create the ansible "workspace" through a reuseable function that the new wrapper script uses). It then runs `packer build` with the correct template. 

As a downside, this does require the packer configuration files to be aware of this special environment they're running in and target the correct path for the playbook and inventory file. While I wish this could be more straight forward, this is probably the price we have to pay for supporting multiple environments and running both plain Ansible as well as packer. 

Marking this as WIP for now as there are some questions to still figure out:
* The common role does a reboot which does not work with packer (the reboot causes a lost ssh connection causing the entire packer process to fail)
* We need to pass the revision done into ansible to checkout the right code.


rylev:
Was able to confirm that the image built with packer is able to successfully run the builder when launched as a fresh instance.

jdno (commented):
Thanks for bootstrapping Packer, @rylev! <3

I think this is a good first step and we can merge it pretty much as-is. In the long-term, I'd like to try to get rid of the `apply` script and its custom logic to handle the environments. Which seems even more benefitial now that Packer requires these hacks as well. Not sure if we can drop the script yet, though...

ansible/roles/common/tasks/apt.yml
```diff
@@ -8,7 +8,7 @@
 - name: install apt packages
   apt:
     name:
-      - aptitude  # needed by ansible itself
+      # - aptitude  # needed by ansible itself
```
jdno, line 11: This seems like an unrelated change. Was this necessary to build the AMI?

packer/README.md
```diff
@@ -2,14 +2,36 @@
...
 
 This directory contains a python script named `packer` which wraps the system `packer`. The script creates an ansible virtual environment and moves the correct ansible configuration in place.
 
 ### Running `packer`
 
-To run the wrapper pass the environment and playbook:
+Before running the packer script, you will need to initialize packer:
+
+```bash
+packer init ./docs-rs-builder
```
jdno, line 22: This needs to be run from within the `packer` directory, correct? If so, do you want to add a `cd packer` or so to the snippet?

packer/docs-rs-builder/builder.pkr.hcl
```diff
@@ -9,6 +9,7 @@ packer {
 
 data "amazon-parameterstore" "revision" {
   name = "/docs-rs/builder/sha"
+  region = "us-east-1"
 }
```
jdno: The formatting is broken here. We run `terraform fmt` in CI, do we want to do the same for Packer?

ansible/ansibleutils.py
```diff
@@ -0,0 +1,44 @@
...
+
+import subprocess
+import pathlib
+import sys
+import shutil
+
+BASE_PATH = pathlib.Path(__file__).resolve().parent
+VENV_PATH = BASE_PATH / ".venv"
+
+# Ansible changes a lot between releases and deprecates a lot of stuff each of
```
meysam81, line 13:
```suggestion
# Ansible changes a lot between releases and deprecates a lot of stuff in each of
```

PR #244, opened 2023-02-16
Allow connecting through aws-psql through an alternate route

rylev:
This is necessary because of two differences in the docs-rs-staging environment and one difference with my current dev machine set up:
* A different bastion host is used 
* A different ssm parameter store key is used to store the database url
* My local user is not the same as the remote user bastion expects

Instead of the old way of calling this script, users would do:

```bash
$ ./aws-psql -i $INSTANCE -d $DATABASE
```

or they can do

```bash
$ ./aws-psql -k $SSM_PARAM_STORE_KEY
```

Additionally, if they want to connect to a different bastion host they can specify that with `-b`

### DIFF
--- a/aws-psql.py
+++ b/aws-psql.py
@@ -5,18 +5,32 @@ import os
 import subprocess
 import sys
 import urllib.parse
+import argparse
 
 
 # Default port to use when connecting to the PostgreSQL database
 DEFAULT_PORT = 5432
 
-# Hostname of the bastion instance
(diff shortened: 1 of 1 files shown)


pietroalbini:
> My local user is not the same as the remote user bastion expects

This can be solved in `~/.ssh/config`:

```
Host bastion.infra.rust-lang.org
    User BASTION_USERNAME
```

PR #250, opened 2023-02-28
Redirect database dump to CloudFront

jdno:
An issue with the database dump for crates was reported on Zulip[^1], and determined to be caused by the file size limit of Fastly. We can currently cache files up to 50MB on Fastly, but the database dump is close to 250MB in size. To avoid this issue, we are redirecting download requests for the dump to CloudFront.

[^1]: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/403s.20from.20.2Fdb-dump.2Etar.2Egz

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/src/config.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/config.rs
@@ -3,6 +3,9 @@ use fastly::ConfigStore;
 // Name of the dictionary. Must match the dictionary in `fastly-static.tf`.
 const DICTIONARY_NAME: &str = "compute_static";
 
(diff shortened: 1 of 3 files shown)


rylev (commented):
Is there anyway to tell if the downstream CDN request fails because the request is too big and then redirect? This would cause cache misses to make it to Fastly but would be more robust.

jdno:
We might be able to do some introspection on the response and figure out if it was caused by the size limit. But that'll require more testing, and my goal here was to unblock sending traffic to Fastly again.

However, I've made a note on my todo list to revisit this at a later time. The code can most certainly be improved and hardened. :+1:

PR #258, opened 2023-03-08
Compress request logs from Fastly

jdno:
Given the volume of traffic and logs, it seems prudent to compress the logs to save on storage and bandwidth costs.


rylev (commented):
Does this mean logs are immediately compressed? Is it possible to compress older logs leaving newer more easily viewable when they are most likely to be needed?

Mark-Simulacrum:
Typically you're not going to view logs in S3 directly, and once you're downloading the file decompressing it is easy. If we're doing any kind of analysis having everything compressed is also convenient.

@jdno Is gzip the only option here? It's what CloudFront uses but if there's better options (e.g. zstd) that might be nice to use here.

jdno:
@Mark-Simulacrum From the [docs](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_compute#compression_codec):

> Valid values are zstd, snappy, and gzip.

It's also possible to tweak the compression level of gzip from 0-9, with 3 being the default. Happy to switch to zstd though if that is the preferred option here.

PR #262, opened 2023-03-12
Implement `sync-team` confirmation

pietroalbini:
**This has already been deployed, but it's not live** (the old automation is still intact in the legacy account).

This PR implements the AWS side of https://github.com/rust-lang/sync-team/pull/48 by copying the `sync-team` deployment into a new AWS Account and creating a Lambda that authenticates the user with GitHub and starts the confirmed sync only after verifying the permission.

I also added a CloudFront distribution at https://confirmation.sync-team-prod.rust-lang.net, since it's slightly easier to remember than https://crbsws6gw5a4rmlblzwlnlahii0epyct.lambda-url.us-east-2.on.aws/.

This PR is best reviewed commit-by-commit, and requires the following PRs to be merged before it's operational:

* https://github.com/rust-lang/sync-team/pull/48
* https://github.com/rust-lang/team/pull/959
* Switching both `rust-lang/sync-team` and `rust-lang/team`'s CIs to interact with the new AWS account (no PR yet).


terragrunt/modules/sync-team/confirmation.tf
```diff
@@ -0,0 +1,132 @@
+locals {
+  domain = "confirmation.sync-team-prod.rust-lang.net"
+}
+
+data "aws_ssm_parameter" "confirmation_parameters" {
+  for_each = toset([
+    "/prod/sync-team-confirmation/github-oauth-client-id",
+    "/prod/sync-team-confirmation/github-oauth-client-secret",
```
rylev, line 8: Hard coding these and the domain seem fine, but we may want to consider passing them as variables.

terragrunt/modules/sync-team/confirmation.tf
```diff
@@ -0,0 +1,132 @@
...
+module "confirmation_certificate" {
+  source  = "../acm-certificate"
+  providers = {
+    aws = aws.us-east-1
+  }
+  domains = [local.domain]
+  legacy  = false
+}
+
+resource "aws_cloudfront_distribution" "confirmation" {
```
rylev, line 69: I'm surprised by wanting an entire cloudfront distribution. Is the reason for this getting the http to https redirect?

terragrunt/modules/sync-team/lambdas/start-sync-team/index.js
```diff
@@ -0,0 +1,8 @@
+const AWS = require('aws-sdk');
```
rylev, line 1: Why are we using JavaScript for this one and Python for the other lambda?

terragrunt/modules/aws-organization/groups.tf
```diff
@@ -111,6 +111,14 @@ locals {
         permissions : [aws_ssoadmin_permission_set.view_only_access] }
       ]
     },
+    # sync-team prod
+    {
+      account : aws_organizations_account.sync_team_prod,
+      groups : [
+        { group : aws_identitystore_group.infra-admins,
```
rylev, line 118: I would like access to this as well if possible.

rylev (commented):
Looks good. I think we should have a readme in the sync-team module describing the architecture at a high-level though.

pietroalbini:
Splitted off the account creation into https://github.com/rust-lang/simpleinfra/pull/290.

terragrunt/modules/aws-organization/groups.tf
pietroalbini: If we're going to implemnent AWS sync (which I'd love to), having admin access to this account would mean escalating privileges to infra-admin levels. It's best to keep perms restricted here.

terragrunt/modules/sync-team/lambdas/start-sync-team/index.js
pietroalbini: I remember spending hours to try and get the Python version to work without success, and switching to js fixed the problem.

terragrunt/modules/sync-team/confirmation.tf
pietroalbini: The cloudfront distribution is needed to have the custom domain, which is less phishable than the random lambda URL.

Mark-Simulacrum: I don't think a CF distribution is a problem, seems pretty lightweight from most perspectives. I think this is fine for now.

terragrunt/modules/aws-organization/groups.tf
Mark-Simulacrum: I think we should grant read-only access (which is safe?) but that can happen separately.

pietroalbini: We would then need to restrict accessing the parameters from the read-only account, and make sure logs are sanitized.

Mark-Simulacrum: Sure. We should probably do that anyway for all humans (i.e., parameters shouldn't be readable by infra-admins either, IMO).

PR #264, opened 2023-03-17
Get metrics from Fastly

jdno:
The [fastly-exporter](https://github.com/fastly/fastly-exporter) provides an easy way to export metrics from Fastly into Prometheus. It is deployed as a stateless container on ECS and scraped by the monitoring server.

### DIFF
--- a/terraform/fastly-exporter/main.tf
+++ b/terraform/fastly-exporter/main.tf
@@ -83,6 +83,8 @@ resource "aws_security_group" "fastly_exporter" {
     protocol  = "tcp"
     # Elastic IP of the monitoring server
     cidr_blocks = ["52.9.166.219/32"]
+    # Load balancer security group, required for health checks
+    security_groups = ["sg-0ddcb954525a46b70"]
   }
 }
 
--- a/terraform/fastly-exporter/main.tf
+++ b/terraform/fastly-exporter/main.tf
@@ -3,7 +3,7 @@ locals {
 }
 
 data "aws_ssm_parameter" "fastly_api_token" {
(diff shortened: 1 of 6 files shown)


jdno:
I'm working remotely today so can't run Ansible to deploy the monitoring server. Will do so after the weekend.

terraform/fastly-exporter/main.tf
```diff
@@ -0,0 +1,109 @@
+locals {
+  name = "fastly-exporter"
+}
+
+data "aws_ssm_parameter" "fastly_api_token" {
+  name = "/prod/crates-io/fastly/api-token"
```
pietroalbini, line 6: Why is this under `/prod/crates-io`?

terraform/fastly-exporter/main.tf
```diff
@@ -0,0 +1,109 @@
...
+  memory = 512
+
+  log_retention_days = 7
+  ecr_repositories_arns = [
+    # This repository does not exist, since we're pulling the imge directly from GitHub.
+    # But the task module cannot be applied without providing at least one ARN here.
+    "arn:aws:ecr:us-west-1:890664054962:repository/fastly-exporter"
+  ]
+
+  containers = <<EOF
```
pietroalbini, line 43: Doesn't matter that much, but it'd be nice to use `jsonencode()` here rather than plaintext JSON.

jdno: We have more blocks like this in Terraform. Let me prepare a separate PR to switch them to `jsonencode`.

PR #267, opened 2023-03-23
Migrate to Global Sign certificates on Fastly

jdno:
OpenSSL < 1.1 (and LibreSSL < 3.2) are not supported by Let's Encrypt, which causes cargo to fail on older operating systems[^1]. To fix this, we are switching `static.crates.io` over to Global Sign certificates. 

[^1]: https://github.com/rust-lang/cargo/issues/11670


jdno:
I've deployed this commit-by-commit on staging and didn't experience any issues. After provisioning the Global Sign certificates, they were _automagically_ used and my macOS 10.13 client was able to download crates. Removing the Let's Encrypt subscription in another `terragrunt apply` worked without issues as well.

PR #270, opened 2023-03-28
Split release module

jdno:
The Terraform module for Rust releases sets up the infrastructure for the release process (`promote-release` tooling) as well as the distribution of releases (CloudFront distributions). These two areas have different access controls, which makes modifying the infrastructure for the distribution of releases more difficult.

This pull request proposes to split the two areas into separate Terraform/Terragrunt modules. The release process will stay in Terraform unchanged, but the CloudFront distributions will be moved into Terragrunt.


Mark-Simulacrum:
This generally seems OK to me, though I don't feel very confident about Terragrunt yet. Seems like we could always iterate further in the future though.

PR #274, opened 2023-04-16
Switch rust-log-analyzer to use S3

pietroalbini:
To be merged after https://github.com/rust-lang/rust-log-analyzer/pull/71.

### DIFF
--- a/terraform/rust-log-analyzer/app.tf
+++ b/terraform/rust-log-analyzer/app.tf
@@ -13,15 +13,12 @@ module "rla" {
   tasks_count      = 1
   platform_version = "1.4.0"
 
-  mount_efs              = "/opt/rla"
-  efs_elastic_throughput = true
-
   environment = {
     CI_REPO     = "rust-lang/rust"
     CI_PROVIDER = "actions"
     EXTRA_ARGS  = "--secondary-repo rust-lang-ci/rust"
     GITHUB_USER = "rust-log-analyzer"
-    INDEX_FILE  = "/opt/rla/rust-lang/rust/actions.idx"
+    INDEX_FILE  = "s3://${aws_s3_bucket.storage.id}/rust-lang/rust/actions.idx"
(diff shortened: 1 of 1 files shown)


terraform/rust-log-analyzer/app.tf
```diff
@@ -13,15 +13,12 @@ module "rla" {
...
-  mount_efs              = "/opt/rla"
-  efs_elastic_throughput = true
-
   environment = {
     CI_REPO     = "rust-lang/rust"
     CI_PROVIDER = "actions"
     EXTRA_ARGS  = "--secondary-repo rust-lang-ci/rust"
     GITHUB_USER = "rust-log-analyzer"
-    INDEX_FILE  = "/opt/rla/rust-lang/rust/actions.idx"
+    INDEX_FILE  = "s3://${aws_s3_bucket.storage.id}/rust-lang/rust/actions.idx"
```
Mark-Simulacrum, line 21: noncurrent_days = 30 is our expiration timeframe which we may want to shrink (256 MB * 24/day * 30 days is ~180 GB, feels a little excessive). But that's still quite cheap so I'm fine either way.

pietroalbini: Note that the index is actually 44 MB. I forgot to look what the other storage was used for in the EFS (and I don't really want to spin up everything to access the EFS again), but I have a good guess on what that might be.

RLA uses the [`atomicwrites`](https://docs.rs/atomicwrites/latest/atomicwrites/) crate to write the file to disk, and that crate works by creating the file first in `.atomicwritesRANDOM/`, and then atomically moving it. I noticed while working on my PR that if the server is shut down when writing to the temporary directory, that directory will be orphaned.

pietroalbini: So in short it might actuually be ~30 GB, which IMO is negligible.

Mark-Simulacrum: Ah, sounds good. Then yeah, let's not worry about this for now.

PR #281, opened 2023-05-31
Add a `project.metadata.docs.rs` redirect in addition to `package.metadata`

jyn514:
Eric Huss says that the two are equivalent in Cargo.toml, so support both in the domain names as well: https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/crate.20files.20with.20unusual.20.60Cargo.2Etoml.60.20files/near/362487043

https://github.com/rust-lang/cargo/blob/58b22f00ef76667ec4f93f4f7936831eaf63f5b7/src/cargo/util/toml/mod.rs#L2030

cc @rust-lang/docs-rs

### DIFF
--- a/terraform/domain-redirects/redirects.tf
+++ b/terraform/domain-redirects/redirects.tf
@@ -102,6 +102,7 @@ module "docs_rs_metadata" {
   to_path = "about/metadata"
   from = [
     "package.metadata.docs.rs",
+    "project.metadata.docs.rs",
   ]
   permanent = true
 }


terraform/domain-redirects/redirects.tf
```diff
@@ -102,6 +102,7 @@ module "docs_rs_metadata" {
   to_path = "about/metadata"
   from = [
     "package.metadata.docs.rs",
+    "project.metadata.docs.rs",
```
epage, line 105: Unsure what the intention of this is but does this elevate the usage of a feature is is legacy / deprecated?   We don't even document this in the [manifest documentation](https://doc.rust-lang.org/cargo/reference/manifest.html) and many of us have gone a long time without even knowing it exists (preparing the PR to merge `cargo add` into `cargo` was when I learned of it after all of my other `cargo add` work).

jyn514:
hmm, fair - I'm ok with closing this if cargo wants to discourage this alias.

PR #282, opened 2023-06-05
Set variables to invalidate releases on Fastly

jdno:
The new environment variables that were added to promote-release in rust-lang/promote-release#66 have been added to the environments. For now, only the dev environment invalidates the cache on Fastly.

### DIFF
--- a/terraform/releases/impl/promote-release.tf
+++ b/terraform/releases/impl/promote-release.tf
@@ -115,6 +115,17 @@ resource "aws_codebuild_project" "promote_release" {
       value = "1"
     }
 
+    environment_variable {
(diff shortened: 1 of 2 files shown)


terraform/releases/impl/promote-release.tf
```diff
@@ -253,6 +264,11 @@ resource "aws_iam_role_policy" "promote_release" {
   })
 }
 
+data "aws_ssm_parameter" "fastly_api_token" {
+  name            = "/${var.name}/promote-release/fastly-api-token"
```
jdno, line 268: The other parameters are all hard-coded to a single `/prod` version. For Fastly, I created two tokens and scoped them to the respective service so that each environment can only interact with its own service.

pietroalbini (approved):

PR #283, opened 2023-06-06
Update ssh keys for Nemo157

Nemo157:
I finally got around to migrating to security key based keys. In case those aren't supported by the servers I've left one non-sk key I can use, but assuming the sk ones work I'll submit another PR to remove that later.

You can confirm these keys are what I'm using for github <https://api.github.com/users/nemo157/keys> (and my commit is signed, this is probably the first time that's ever actually been useful 😁).


jdno:
I'll deploy this in the coming hour. 🚀

PR #286, opened 2023-06-07
Add podman to dev desktop

fee1-dead:
per [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Cloud.20desktop.20package.20request)

### DIFF
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -51,6 +51,7 @@
       - lld
       - lldb
       - mold
+      - podman
     state: present
 
 - name: Install tools for x86


albertlarsan68:
I think there may be more steps than just adding podman to the apt install list.

fee1-dead:
> I think there may be more steps than just adding podman to the apt install list.

Per the [installation instructions](https://podman.io/docs/installation) on their site, I don't see additional steps required other than installing from the package manager.

albertlarsan68:
Per [this other official doc page](https://github.com/containers/podman/blob/main/docs/tutorials/rootless_tutorial.md), it requires a bit of setup to make sure non-root users have access.

fee1-dead:
Ah I see. Then I don't think I can write an ansible config for that since I don't have a lot of experience in ansible.

jdno:
Thanks for the links. 🙂

I had a brief look at the installation guide and the only thing that stands out as a potential issue is reserving UIDs in `/etc/subuid`. Since we create users dynamically from rust-lang/team, this might get annoying to automate.

I'll try to make some time either tomorrow or next week to do look more into this.

Nilstrieb:
@jdno friendly ping in case you forgot about it :)

jdno:
Oh my time is flying 😱 I haven't forgotten about it, but I also need to block some dedicated time next week. 🤦‍♂️

jdno:
I've installed Podman on `dev-desktop-staging`. Still need to do a few more tests, but feel free to poke it as well and let me know if you find any issues.

jdno:
I've created a new pull request for my branch that installs Podman and configures the rootless mode here: https://github.com/rust-lang/simpleinfra/pull/320. I'll give folks a chance to review it, and plan to deploy it to production after the weekend.

jdno:
I've deployed the other PR to the dev-desktops last week, so Podman should be available and usable for everyone. If that's not the case, let me know. Otherwise we can probably close this PR. :)

PR #289, opened 2023-06-09
Create weighted DNS records for Rust releases

jdno:
Weighted DNS records allow us to route traffic dynamically between CloudFront and Fastly. The existing DNS record is being modified, preventing any issues from removing and then recreating it so that the migration can be performed without downtime.

### DIFF
--- a/terragrunt/accounts/legacy/releases-dev/release-distribution/terragrunt.hcl
+++ b/terragrunt/accounts/legacy/releases-dev/release-distribution/terragrunt.hcl
@@ -18,6 +18,6 @@ inputs = {
 
   static_ttl = 86400 // 1 day
 
-  static_cloudfront_weight = 100
-  static_fastly_weight = 0
+  static_cloudfront_weight = 50
+  static_fastly_weight = 50
 }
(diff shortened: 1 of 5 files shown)


Mark-Simulacrum (approved):

terragrunt/accounts/legacy/releases-dev/release-distribution/terragrunt.hcl
```diff
@@ -15,4 +15,7 @@ inputs = {
 
   static_bucket = "dev-static-rust-lang-org"
   log_bucket = "rust-release-logs"
+
+  static_cloudfront_weight = 0
```
Mark-Simulacrum, line 19: We may want to move this back to 50/50 or something, since otherwise we're not testing that CF is working on dev-static.

jdno: I'm afraid that realistically the traffic volume against `dev-static` isn't high enough to have a lot of confidence whether both are working anyways. We'll have to test changes against both CDNs manually whenever we make them. But I'll drop this to 50/50 and to the same for `static.staging.crates.io`.

Mark-Simulacrum: Yeah, that's true. I think some amount of passive testing is likely though (e.g., when we announce a pre-release). That will help with changes upstream that aren't caused by us - things like the artifact size limitations we ran into previously - while being low-cost on our end.

PR #293, opened 2023-06-12
Enable cache invalidations on Fastly in production

jdno:
Cache invalidations on Fastly are gated behind a feature flag, which this change enables. Invalidations require the static domain name as well as an API key, both of which have been added to the service in a previous commit.


jdno:
I haven't deployed this yet by the way.

PR #313, opened 2023-07-04
Rewrite URLs with plus character for crates.io

jdno:
The CDNs for `static.crates.io` now rewrites URLs that contain a plus character to a properly URL-encoded version. This ensures that clients access objects in S3 in a consistent way, and avoids any issues with Amazon S3's special handling of the plus character.

See https://github.com/rust-lang/crates.io/issues/4891 for more details.

### DIFF
--- a/terragrunt/modules/crates-io/cloudfront-static.tf
+++ b/terragrunt/modules/crates-io/cloudfront-static.tf
@@ -4,22 +4,11 @@ locals {
   cloudfront_domain_name = "cloudfront-${var.static_domain_name}"
 }
 
-data "aws_iam_role" "cloudfront_lambda" {
-  name = "cloudfront-lambda"
-}
-
-module "lambda_static" {
-  source = "../aws-lambda"
-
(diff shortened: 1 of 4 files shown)


terragrunt/modules/crates-io/cloudfront-static.tf
```diff
@@ -45,6 +63,12 @@ resource "aws_cloudfront_distribution" "static" {
         forward = "none"
       }
     }
+
+    lambda_function_association {
```
Mark-Simulacrum, line 67: Based on last ~30 days I think this would mean 339 million * $0.60/million requests - so roughly an extra ~$200/month. I think if we can switch to cloudfront functions we get 6x cheaper pricing ($0.10/million requests) and presumably roughly the same functionality for the redirects desired here?

jdno: Great suggestion! I refactored the code to use a CloudFront function instead.

PR #317, opened 2023-07-13
Monitor EC2 hosts with Datadog agent

jdno:
The Datadog agent collects events and metrics from hosts and sends them to Datadog[^1]. It is deployed to our hosts using an Ansible role provided by Datadog and distributed through Ansible Galaxy[^2]. The `apply` script has been modified to install dependencies from Ansible Galaxy before running a playbook.

[^1]: https://docs.datadoghq.com/agent/
[^2]: https://galaxy.ansible.com/Datadog/datadog

### DIFF
--- a/ansible/envs/staging/group_vars/all.yml
+++ b/ansible/envs/staging/group_vars/all.yml
@@ -1,5 +1,10 @@
 ---
 
+# Fetch all the group params from AWS SSM
+ssm_all: "{{ lookup('aws_ssm', '/staging/ansible/all/', region='us-west-1', shortnames=true, bypath=true, recursive=true) }}"
+
+vars_datadog_api_key: "{{ ssm_all['datadog-api-key'] }}"
+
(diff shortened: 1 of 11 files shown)


jdno:
I've set this up only for the dev-desktops, but if there are no concerns with this approach I'd like to add this to the other hosts as well.

ansible/envs/staging/group_vars/all.yml
```diff
@@ -1,5 +1,10 @@
 ---
 
+# Fetch all the group params from AWS SSM
+ssm_all: "{{ lookup('aws_ssm', '/staging/ansible/all/', region='us-west-1', shortnames=true, bypath=true, recursive=true) }}"
```
jdno, line 4: If this becomes an issue, because people without access to SSM want to deploy staging services, we can just move this into the group vars for the dev-desktops. I want to deploy the agent on staging _somewhere_, but it doesn't have to be on all machines necessarily.

PR #318, opened 2023-07-13
Manage users on Datadog with Terraform

jdno:
A Terraform module has been created similar to the existing modules to manage team access to Datadog.


terraform/team-members-datadog/.terraform.lock.hcl
```diff
@@ -0,0 +1,24 @@
+# This file is maintained automatically by "terraform init".
```
meysam81, line 1: Shouldn't you ignore this file, though?

terraform/team-members-datadog/_terraform.tf
```diff
@@ -0,0 +1,22 @@
+// Configuration for Terraform itself.
+
+terraform {
+  required_version = "~> 1"
```
meysam81, line 4: Isn't it a better idea to pin it to something before the BUSL license was introduced? e.g. `~> 1.5`

terraform/team-members-datadog/_terraform.tf
```diff
@@ -0,0 +1,22 @@
...
+
+  required_providers {
+    datadog = {
+      source  = "datadog/datadog"
+      version = "3.24.1"
```
meysam81, line 9: And perhaps pinning this version to include all the patch ones: `~> 3.24`

terraform/team-members-datadog/_terraform.tf
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  backend "s3" {
+    bucket         = "rust-terraform"
+    key            = "simpleinfra/team-members-datadog.tfstate"
+    region         = "us-west-1"
+    dynamodb_table = "terraform-state-lock"
+    encrypt        = true
+  }
+}
```
meysam81, line 20: I wander, doesn't `simpleinfra` use Terragrunt? Wouldn't it be better to use the DRY it offers?

terraform/team-members-datadog/.terraform.lock.hcl
jdno: No, we want to make sure that we're all using the same versions. `terraform init` fetches the versions that are locked in `.terraform.lock.hcl`, and only upgrades them when explicitly called with `terraform init -upgrade`.

terraform/team-members-datadog/_terraform.tf
jdno: AFAIK the license change doesn't impact us as end-users of Terraform, since we're not embedding Terraform or its source code in a commercial product that we're selling. At least that's how I understand their [FAQ](https://www.hashicorp.com/license-faq#implications-of-change-for-users).

terraform/team-members-datadog/_terraform.tf
jdno: There's a broader discussion coming up once in a while about our infrastructure-as-code tooling and its issues and limitations. One of the things that we're struggling with is keeping our modules and dependencies up-to-date. It's not too bad here, but for other modules it's quite a lot of work to update something, because then every service that depends on that module needs to be manually redeployed.

I think it's fine to leave it as-is for now, since we really need to solve this in an holistic way for the whole repository.

terraform/team-members-datadog/_terraform.tf
jdno: Terragrunt is really handy for things that have multiple environments like `production` and `staging`. The team access will only ever have a single instance, so we haven't bothered moving it over to Terragrunt yet. And since the other rest of the team access is managed with Terraform as well, I didn't want to make this a one-off configuration in a different place.

PR #330, opened 2023-07-24
Add alert when there are too many queued jobs on crates.io

pietroalbini:
cc @rust-lang/crates-io

### DIFF
--- a/ansible/playbooks/monitoring.yml
+++ b/ansible/playbooks/monitoring.yml
@@ -188,6 +188,12 @@
               labels:
                 dispatch: pagerduty-cratesio
 
+            - alert: TooManyQueuedJobs
+              expr: cratesio_service_background_jobs{env="prod"} > 100
+              for: 1m
+              labels:
+                dispatch: pagerduty-cratesio
+
         - name: docsrs
           rules:
             - alert: DocsRsDown


ansible/playbooks/monitoring.yml
```diff
@@ -188,6 +188,12 @@
               labels:
                 dispatch: pagerduty-cratesio
 
+            - alert: TooManyQueuedJobs
+              expr: cratesio_service_background_jobs{env="prod"} > 100
```
Turbo87, line 192: we introduced a `priority` column for background jobs lately and when we backfilled the `rust-version` values a few weeks ago we had several thousand low-priority background jobs queued up. I guess such a case would trigger the pagerduty alert if we merged it like this. Maybe we should somehow feed the priority into the monitoring system too? For the `bin/monitor` application we only consider `priority >= 0` (normal or higher).

pietroalbini: I don't think the priority is exposed through the `cratesio_service_background_jobs` metric though. We should add it first.

pietroalbini: Will update once https://github.com/rust-lang/crates.io/pull/6871 is deployed.

Turbo87: the PR is deployed in staging and production now

PR #335, opened 2023-07-29
feat: allow users to select their shell

### LINKED ISSUE #333: Enable users to choose their shell on dev-desktops
A recurring feature request for the dev-desktops has been the ability to customize a user's default shell. The following message in Zulip and the following discussion is a good example: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Cloud.20compute.20.2F.20Dev.20desktop.20feedback/near/307585274.

## Proposed Design

_The following is only a proposal and the final implementation may look different._

The simplest solution is probably to add a new configuration to our [Ansible role for the dev-desktops](https://github.com/rust-lang/simpleinfra/tree/master/ansible/roles/dev-desktop) that allows users to opt-in to an alternative shell:

```yaml
# defaults/main.yml
vars_shells:
  gh-jdno: /usr/bin/fish
  gh-octocat: /usr/bin/zsh
```

The role should then loop over the users in the list, check whether the user actually exists, and if so set their default shell to the selected shell from the configuration file.

This allows users to create a pull request against this repository to change their shell. An admin still has to review the pull request and run Ansible, but this allows us to confirm that the user chose a valid shell.

It's be great to add user-facing documentation on how to change the shell to https://forge.rust-lang.org/infra/docs/dev-desktop.html (in the [rust-lang/forge](https://github.com/rust-lang/rust-forge) repository).

## Alternatives

The discussion linked at the beginning of this issue proposes a solution based on the [rust-lang/team](https://github.com/rust-lang/team) repository and a new configuration block in a user's configuration file. While it would remove the need for admins to apply the configuration on the dev-desktops, this aproach would require significant changes to the schema of the team repository and the `team_login` binary that creates the user accounts on the dev-desktops.

This seems too complicated and time-consuming compared to a simpler solution with Ansible, given that the request to change a shell does not happen very frequently.

meysam81:
fixes #333


jdno (changes requested):
The `sudo_users` and `unprivileged_users` are just the administrators. The end-users of the dev-desktops are created by a Rust binary that you can find inside [ansible/roles/dev-desktop/files/team_login](https://github.com/rust-lang/simpleinfra/tree/master/ansible/roles/dev-desktop/files/team_login). Probably only end-users will want to change their shell.

Here's a Stack Overflow question that uses the approach that I would've taken: https://serverfault.com/questions/1001235/how-can-change-shell-for-a-user-via-ansible. Sadly Ansible does not support looping over blocks, but this seems to be the second best option.

We should then be able to loop over the entries in `vars_user_shell` and check if the user exists on the system, and if so set their shell using the `ansible.builtin.user` module.

meysam81:
@jdno how about this one? :)

jdno (commented):
Yes! That's exactly what I had in mind. 🙂

Can we remove the `usermod.yml` file, since that only calls `set_shell.yml`?

And out of curiosity, did you try to run this against a local development environment?

meysam81:
@jdno 
Thanks for taking the time to review this.

The changes were applied and it seemed appropriate to keep the `usermod` name for any future change to the linux user.
I also tested the playbook locally and it populates correctly:

```plaintext
$ tail -n 3 /etc/passwd
rylev:x:1001:1001::/home/rylev:/bin/sh
meysam:x:1002:1002::/home/meysam:/usr/bin/zsh
jdno:x:1003:1003::/home/jdno:/usr/bin/fish
```

ansible/roles/dev-desktop/tasks/usermod.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Check user {{ item.username }} exists
+  ansible.builtin.getent:
+    database: passwd
+    key: "{{ item.username }}"
+  register: user_exists
+  failed_when: false
```
jdno, line 7: I tested this on `staging` and think that this needs to be:

```suggestion
  ignore_errors: yes
```

Otherwise the playbook will continue with the `Set the user shell` task and create the missing user.

jdno (changes requested):

ansible/roles/dev-desktop/tasks/usermod.yml
meysam81: yes, you're absolutely right

meysam81:
@jdno 
All yours now.

Let me know if you need any other change :pray:

jdno (approved):
Thanks, @meysam81! This works great on `staging`. I'm going to merge and deploy this. 🥳

PR #336, opened 2023-08-02
Deploy custom role for Wiz integration

jdno:
Wiz provides a managed AWS role for its integration. The role grants Wiz access to certain S3 buckets that we want to exclude from its analysis. Namely, buckets contain Terraform state files and the Rust release keys. The provided AWS role has been copied into the project and modified to deny access to those buckets.

### DIFF
--- a/terragrunt/modules/wiz/main.tf
+++ b/terragrunt/modules/wiz/main.tf
@@ -121,8 +121,6 @@ resource "aws_iam_role_policy" "tf-policy" {
         "Action" : [
           "ec2:CopySnapshot",
           "ec2:CreateSnapshot",
-          "kms:CreateKey",
-          "kms:DescribeKey",
           "ec2:GetEbsEncryptionByDefault",
           "ec2:DescribeSnapshots"
         ],
@@ -143,19 +141,6 @@ resource "aws_iam_role_policy" "tf-policy" {
           "arn:aws:kms:*:*:alias/wizKey",
           "arn:aws:kms:*:*:key/*"
         ]
-      },
-      {
-        "Action" : [
-          "kms:CreateGrant",
-          "kms:ReEncryptFrom"
-        ],
-        "Condition" : {
-          "StringLike" : {
-            "kms:ViaService" : "ec2.*.amazonaws.com"
-          }
-        },
-        "Effect" : "Allow",
-        "Resource" : "*"
       },
       {
         "Action" : [
--- a/terragrunt/modules/wiz/main.tf
+++ b/terragrunt/modules/wiz/main.tf
@@ -170,16 +170,11 @@ resource "aws_iam_role_policy" "tf-policy" {
       },
       {
         "Action" : [
-          "s3:*",
+          "s3:GetObject"
         ],
         "Effect" : "Deny",
         "Resource" : [
-          "arn:aws:s3:::*terraform*",
-          "arn:aws:s3:::*tfstate*",
-          "arn:aws:s3:::*tf?state*",
-          "arn:aws:s3:::*cloudtrail*",
-          "arn:aws:s3:::elasticbeanstalk-*",
-          "arn:aws:s3:::rust-release-keys",
+          "*"
         ],
         "Sid" : "WizAccessS3"
       }
--- a/terragrunt/modules/wiz/main.tf
+++ b/terragrunt/modules/wiz/main.tf
@@ -1,6 +1,205 @@
-module "wiz" {
-  source        = "https://s3-us-east-2.amazonaws.com/wizio-public/deployment-v2/aws/wiz-aws-native-terraform-terraform-module.zip"
-  remote-arn    = "arn:aws:iam::830522659852:role/prod-us43-AssumeRoleDelegator"
-  external-id   = "fc959d31-537c-4108-8b87-9af9e0c9b8d2"
-  data-scanning = false
+resource "aws_iam_role_policy" "tf-policy" {
+  name = "WizCustomPolicy"
+  role = aws_iam_role.user-role-tf.id
+
+  policy = jsonencode({
+    "Statement" : [
+      {
+        "Action" : [
+          "acm:GetCertificate",
+          "apigateway:GET",
+          "backup:DescribeGlobalSettings",
+          "backup:GetBackupVaultAccessPolicy",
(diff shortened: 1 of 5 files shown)


terragrunt/modules/wiz/main.tf
```diff
@@ -1,6 +1,205 @@
...
+          "cloudwatch:GetMetricStatistics"
+        ],
+        "Effect" : "Allow",
+        "Resource" : "*"
+      },
+      {
+        "Action" : [
+          "ec2:CopySnapshot",
+          "ec2:CreateSnapshot",
+          "kms:CreateKey",
```
Mark-Simulacrum, line 124: Why does Wiz need to create KMS keys?

terragrunt/modules/wiz/main.tf
```diff
@@ -1,6 +1,205 @@
...
+        "Action" : "kms:CreateAlias",
+        "Effect" : "Allow",
+        "Resource" : [
+          "arn:aws:kms:*:*:alias/wizKey",
+          "arn:aws:kms:*:*:key/*"
+        ]
+      },
+      {
+        "Action" : [
+          "kms:CreateGrant",
```
Mark-Simulacrum, line 149: This seems like it allows Wiz to effectively bypass any KMS policies we might establish? We don't today, but if we e.g. managed signing keys as a kms resource I think this would allow wiz to sign things?

terragrunt/modules/wiz/main.tf
```diff
@@ -1,6 +1,205 @@
...
+          "s3:*",
+        ],
+        "Effect" : "Deny",
+        "Resource" : [
+          "arn:aws:s3:::*terraform*",
+          "arn:aws:s3:::*tfstate*",
+          "arn:aws:s3:::*tf?state*",
+          "arn:aws:s3:::*cloudtrail*",
+          "arn:aws:s3:::elasticbeanstalk-*",
+          "arn:aws:s3:::rust-release-keys",
```
Mark-Simulacrum, line 197: Can we allowlist instead or in addition to denylist, with an alarm/audit to make sure all buckets are listed?

terragrunt/modules/wiz/main.tf
walterhpearce: Since we don't use KMS right now, lets just drop these permissions. We can revisit explicit allowances if we ever are worried about vulnerabilities in that aspect (Hint: I don't think we ever will be)

terragrunt/modules/wiz/main.tf
walterhpearce: I'd argue with an allow list we would be losing benefit we get from Wiz s3 monitoring, which is getting insight into new buckets and making sure they were established correctly. I am more concerned about a bucket being stood up and used with unintentionally incorrect settings/permissions out-of-band than settings being changed on an existing bucket we know about (although that is also a concern). 

Are you suggesting we have a out-of-band job, maybe a validation lint here in `simpleinfra`, that confirms that the allowlist matches all buckets currently in existence mod explicit denials?

Mark-Simulacrum: Yes, I was under the impression that Wiz already is doing that validation - basically, I am 100% onboard with every bucket having an explicit decision, but I want that decision, at least while we still have buckets with sensitive contents. If we e.g. switch to using KMS encryption for terraform state and any other secrets in s3 rather than just s3 key encryption, then we wouldn't need this (can just not grant KMS key access so the objects aren't interesting).

jdno:
I've removed the KMS actions that were flagged above. 👍

terragrunt/modules/wiz/main.tf
jdno: If I'm not missing anything, the only actions that Wiz is allowed to take on S3 are:

- `s3:GetBucketNotification`
- `s3:GetMultiRegionAccessPointPolicy`
- `s3:ListMultiRegionAccessPoints`

Those seem fine to have enabled for all buckets, even sensitive ones we create accidentally. Personally, I don't think it's worth the added complexity to maintain an allow list manually and create a script/GitHub Action that periodically checks the list against our AWS accounts.

Mark-Simulacrum: Now I'm confused, because those don't seem like enough to even get a list of buckets or check any of the policies I'd expect Wiz to want to check? Are we sure these permissions are a complete list?

jdno: I totally missed that the AWS-managed [SecurityAudit](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/SecurityAudit.html) group also includes a bunch of permissions for S3.

```
"s3:GetAccelerateConfiguration",
"s3:GetAccessPoint",
"s3:GetAccessPointPolicy",
"s3:GetAccessPointPolicyStatus",
"s3:GetAccountPublicAccessBlock",
"s3:GetAnalyticsConfiguration",
"s3:GetBucket*",
"s3:GetEncryptionConfiguration",
"s3:GetInventoryConfiguration",
"s3:GetLifecycleConfiguration",
"s3:GetMetricsConfiguration",
"s3:GetObjectAcl",
"s3:GetObjectVersionAcl",
"s3:GetReplicationConfiguration",
"s3:ListAccessPoints",
"s3:ListAllMyBuckets",
```

Mark-Simulacrum: Okay, in that case it looks like Wiz isn't even requesting GetObject permissions -- maybe the right thing to do is for us to explicitly deny that permission and otherwise not need a deny list on these buckets?

jdno: The Wiz-provided module had another resource that would request `GetObject` _if_ data scanning was enabled. But I removed that resource when copying over the policies, since we're not planning on enabled the feature anyways.

You mean changing the last block to something like this:

```json
{
  "Action": [
    "s3:GetObject"
  ],
  "Effect": "Deny",
  "Resource": [
    "*"
  ],
  "Sid": "WizAccessS3"
}
```

Mark-Simulacrum: Yes, exactly. I think that covers our bases and avoids needing bucket name matches (which could diverge accidentally).

jdno: Done.

PR #345, opened 2023-09-11
Increase block size for segmented caching

jdno:
Increasing the block size for segmented caching results in much better performance, but might cause segments to be evicted from the cache earlier and thus lowering the cache hit rate. Fastly does not recommend raising the block size above 10MB for this exact reason. We are starting at this limit to optimize for performance, and might lower the size if we observe issues with the cache hit rate.

Read more: https://developer.fastly.com/reference/vcl/variables/segmented-caching/segmented-caching-block-size/

This change also enables the streaming miss feature, which improves the time to first-byte. We've enabled it while investigating the slow download speeds.

Read more: https://docs.fastly.com/en/guides/streaming-miss


jdno:
I've deployed this to both staging and production and am watching the metrics closely to ensure everything works smoothly. The change in block sizes busted the cache, but the cache hit rate is slowly returning to normal levels.

Mark-Simulacrum (approved):

meysam81:
Is there any public Grafana to see the result of this change? :thinking:

jdno:
> Is there any public Grafana to see the result of this change? 🤔

No. I checked this in Fastly's dashboard. We theoretically send the metrics to Prometheus as well, but we haven't created a dashboard for it yet.

PR #348, opened 2023-09-23
feat: add upload-metrics-artifact GitHub Action ✨

meysam81:
fixes rust-lang/infra-team#74

You can see the successful test-run [here][run].

[run]: https://github.com/meysam81/simpleinfra/actions/runs/6283641163

### DIFF
--- a/github-actions/upload-metrics-artifact/README.md
+++ b/github-actions/upload-metrics-artifact/README.md
@@ -1,0 +1,45 @@
+# Upload Metrics Artifact
+
+This GitHub Action will upload metrics collected by CI run to GitHub Artifacts.
+This allows for future runners or other actions to download the artifact and
+upload them to S3 bucket that is used for monitoring GitHub CI.
+
+You can follow the discussion in [the relevant GitHub issue][issue-source].
+
+## Usage
+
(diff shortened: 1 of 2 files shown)


meysam81:
@jdno

Please take a look and let me know if there's anything missing.

meysam81:
On second thought, I think we can also make this action more generic with the following `action.yml` definition.

```yaml
name: Collect CI metrics
description: Collect CI metrics and upload them to GitHub Artifacts

inputs:
  collect-metrics:
    description: Whether to collect metrics or not
    required: false
    default: "1"
  metrics-filepath:
    description: Comma or space separated list of metrics files to be uploaded to GitHub artifact
    required: true
    default: build/cpu-usage.csv
  upload-artifacts:
    description: Whether to upload artifacts or not
    required: false
    default: "1"

runs:
  using: node20
  pre-if: inputs.collect-metrics == "1"
  pre: collect.js
  main: index.js
  post-if: inputs.upload-artifacts == "1"
  post: upload.js
```

Please let me know if this is a better approach.


Which, then, requires only one step in a job definition instead of two if we go with the first draft of this PR.

Issue #338, opened 2023-08-07
Retry renewal of SSL certificates


jdno:
The `renew-ssl-certs.service` fails periodically, which triggers the following alert in Grafana:

```
The systemd unit renew-ssl-certs.service on the <server>:9100 instance failed to execute.
```

The fix for this alert is to restart the service manually:

```shell
sudo systemctl restart renew-ssl-certs.service
```

We should investigate the reason why the service fails in the first place and retry automatically.

Mark-Simulacrum:
Relevant logs:

```
May 23 00:00:01 monitoring.infra.rust-lang.org systemd[1]: Starting Renew SSL certificates...
May 23 00:00:03 monitoring.infra.rust-lang.org renew-ssl-certs[2866]: 2022/05/23 00:00:03 [INFO] [grafana.rust-lang.org] acme: Trying renewal with 647 hours remaining
May 23 00:00:03 monitoring.infra.rust-lang.org renew-ssl-certs[2866]: 2022/05/23 00:00:03 [INFO] [grafana.rust-lang.org] acme: Obtaining bundled SAN certificate
May 23 00:00:09 monitoring.infra.rust-lang.org renew-ssl-certs[2866]: 2022/05/23 00:00:09 acme: error: 500 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-order :: urn:ietf:params:acme:error:serverInternal :: Error creating new order, url:
May 23 00:00:09 monitoring.infra.rust-lang.org systemd[1]: renew-ssl-certs.service: Main process exited, code=exited, status=1/FAILURE
May 23 00:00:09 monitoring.infra.rust-lang.org systemd[1]: renew-ssl-certs.service: Failed with result 'exit-code'.
May 23 00:00:09 monitoring.infra.rust-lang.org systemd[1]: Failed to start Renew SSL certificates.
```

My guess is this is an upstream spurious failure of some kind -- presumably one-off and resolvable by retrying. Let's Encrypt doesn't note any problems on their status page at this time:

<img width="616" alt="image" src="https://github.com/rust-lang/simpleinfra/assets/5047365/0af3b7ab-026a-4000-8771-a012d63f80cd">

So my best guess is low levels of fuzz in their availability; we likely can fix this by retrying on our side at the systemd layer or within the executed script.

jdno:
We haven't seen this error in a while, so closing the issue for now.

PR #349, opened 2023-09-23
fix: retry failed ssl renewal with jittered backoff 🐛

meysam81:
fixes rust-lang/simpleinfra#338

For your reference, you can see the templated file below:

<details>

```bash
#!/bin/bash
#
# This file is managed by Ansible. Only change it through the rust-lang/simpleinfra repository.
#
set -euo pipefail
IFS=$'\n\t'

# During the initial renew nginx will fail to run as the certificate is
# missing, so lego will have to serve port 80.
if [[ $# -eq 1 ]] && [[ $1 = "initial-renew" ]]; then
    renew_kind="--http.port=:80"
    action="run"
else
    renew_kind="--http.webroot=/var/run/acme-challenges"
    action="renew"
fi

# Use the local Pebble instance to get a dummy certificate
server="https://localhost:14000/dir"
# Use Pebble's CA to validate connections to it
export LEGO_CA_CERTIFICATES=/etc/pebble/certs/pebble.minica.pem
# Remove the account as Pebble doesn't persist it
rm -rf "/etc/ssl/letsencrypt/accounts/localhost_14000/admin@rust-lang.org"

retries="10"
wait_time="5"

lego_cmd="lego --email 'admin@rust-lang.org' \
        --server '${server}' \
        --accept-tos \
        --path /etc/ssl/letsencrypt \
        --http \
        ${renew_kind} \
        -d 'dummy' \
        ${action}"

function run_with_retries {
    command=$1
    local i=0
    while true; do
        ${command}
        exit_code=$?

        if [ ${exit_code} -eq 0 ]; then
            break
        fi

        if [ ${i} -ge ${retries} ]; then
            exit ${exit_code}
        fi

        jitter=$(($RANDOM % 10))
        wait_time_with_jitter=$((${wait_time} + ${jitter}))

        echo "Command failed with exit code ${exit_code}. Retrying in ${wait_time_with_jitter} seconds..."
        sleep ${wait_time_with_jitter}

        i=$(($i + 1))
    done
}

set +e
run_with_retries $lego_cmd
set -e

sudo /etc/ssl/letsencrypt/after-renew
```
</details>

Running the executable on the machine gave me the output you can see in the screenshot below.

failing-but-retrying-with-jitter-image

### DIFF
--- a/ansible/roles/letsencrypt/defaults/main.yml
+++ b/ansible/roles/letsencrypt/defaults/main.yml
@@ -1,3 +1,6 @@
 ---
 
 dummy_certs: false
+
+vars_ssl_renew_number_of_retries: 10
+vars_ssl_renew_retry_delay: 5
--- a/ansible/roles/letsencrypt/templates/renew-ssl-certs.sh
+++ b/ansible/roles/letsencrypt/templates/renew-ssl-certs.sh
@@ -65,6 +65,8 @@ function run_with_retries {
     done
 }
 
+set +e
 run_with_retries $lego_cmd
+set -e
 
 sudo /etc/ssl/letsencrypt/after-renew
--- a/ansible/roles/letsencrypt/templates/renew-ssl-certs.sh
+++ b/ansible/roles/letsencrypt/templates/renew-ssl-certs.sh
@@ -26,15 +26,45 @@ rm -rf "/etc/ssl/letsencrypt/accounts/localhost_14000/{{ email }}"
 server="https://acme-v02.api.letsencrypt.org/directory"
 {% endif %}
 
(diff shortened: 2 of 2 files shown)


meysam81:
@jdno Please have a look and let me know if there's anything missing :pray:

Mark-Simulacrum:
Can we structure this implementation to use retries at the systemd level (https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/letsencrypt/templates/renew-ssl-certs.service)? I feel like that might be less ad-hoc in terms of implementation and a little cleaner than the current solution.

shepmaster:
Superseded by #393

PR #356, opened 2023-10-23
Pass Accept-Encoding header to origin

jdno:
The Accept-Encoding header is now passed to the origin so that it can respond with compressed responses.

This fixes an inconsistency with the backend, which prefers Brotli compression while CloudFront prefers gzip. Because the header is now passed to the origin, the Brotli-compressed response from the origin is passed through to the client. Previously, CloudFront would fetch uncompressed resources and compress them on the fly using gzip.

cc @Turbo87


Turbo87 (approved):
Note that this does not apply to the `/assets` behavior, since we discovered that this behavior never worked because `/assets` refers to a fixed path, while `/assets/*` would have been the correct path pattern... 😅 

Passing the header for the regular behavior is still useful for `robots.txt` and other top-level files though.

PR #357, opened 2023-10-23
Remove unused cache behavior

jdno:
The custom cache behavior for `/assets` was never used, since the pattern matches only the specific `/assets` URL and not any resources served under it (e.g. `/assets/foo`). This was discovered while investigating the inconsistencies in #356.

Since the configuration was never used in the first place, we are removing it to remove some complexity.

cc @Turbo87

### DIFF
--- a/terragrunt/modules/crates-io/cloudfront-webapp.tf
+++ b/terragrunt/modules/crates-io/cloudfront-webapp.tf
@@ -54,24 +54,6 @@ resource "aws_cloudfront_distribution" "webapp" {
     }
   }
 
-  // Cache assets ignoring headers, query strings and cookies.
-  ordered_cache_behavior {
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):

Mark-Simulacrum (approved):
I'm fine with removing but it also seems like plausibly useful to customize some parts for it. In any case, seems okay for now.

Turbo87:
@Mark-Simulacrum agreed, though it is apparently still using some sort of legacy caching configuration for CloudFront. we thought it was probably better to recreate it using their (new?) recommended caching policy system instead, and potentially wait for crates.io to serve assets directly from S3 instead of Heroku. simplifying the config here for now will likely make the migration easier :)

PR #362, opened 2023-11-10
Set my default shell to fish on dev-desktops

thomcc:
Thanks.


thomcc:
Is there any way the version could be updated too? My personal fish configuration and scripts don't work with fish 3.3.1 which seems to be installed (which is from 2021). I need at least 3.4.0 ([which added new interpolation syntax](https://fishshell.com/docs/current/relnotes.html#fish-3-4-0-released-march-12-2022)), ideally newer since that's still over a year old.

jdno:
We used to install fish from Ubuntu's default package repository, but I've switched to fish's PPA and installed the latest version from there. :+1:

jdno (approved):

jdno:
This has now been deployed to the dev-desktops. 🚀

PR #365, opened 2023-11-17
crates.io: Add support for download URLs without placeholders

Turbo87:
In https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/placeholders.20in.20index.20config.2Ejson I asked what cargo version added support for placeholders in the `dl` config field of the package index and apparently the answer is Rust 1.24.

In the future the crates.io team would like to move crate file downloads directly to the CDN, instead of them having to go through our own backend, since that has proven to be a somewhat problematic bottleneck. To be able to support earlier cargo versions, this PR introduces URL rewriting for `/crates/:crate/:version/download` requests for the CloudFront and Fastly CDNs.

This should allow us to change https://github.com/rust-lang/crates.io-index/blob/27423df2d3aecc0642d64eee2b321ffa1e6860ba/config.json#L2 to `"dl": "https://static.crates.io/crates"` in the future.

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/Cargo.toml
+++ b/terragrunt/modules/crates-io/compute-static/Cargo.toml
@@ -19,6 +19,8 @@ derive_builder = "0.12.0"
 fastly = "0.9.1"
 log = "0.4.17"
 log-fastly = "0.9.1"
+once_cell = "1.18.0"
+regex = "1.7.1"
 serde = { version = "1.0.152", features = ["derive"] }
 serde_json = "1.0.92"
 time = { version = "0.3.17", features = ["serde-human-readable"] }
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -2,6 +2,8 @@ use fastly::http::{Method, StatusCode};
 use fastly::{Error, Request, Response};
 use log::{info, warn, LevelFilter};
 use log_fastly::Logger;
+use once_cell::sync::Lazy;
+use regex::Regex;
 use serde_json::json;
 use time::OffsetDateTime;
 
@@ -74,6 +76,7 @@ fn handle_request(config: &Config, mut request: Request) -> Result<Response, Err
 
     set_ttl(config, &mut request);
     rewrite_urls_with_plus_character(&mut request);
+    rewrite_download_urls(&mut request);
 
     // Database dump is too big to cache on Fastly
     if request.get_url_str().ends_with("db-dump.tar.gz") {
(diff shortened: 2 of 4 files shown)


terragrunt/modules/crates-io/compute-static/Cargo.toml
```diff
@@ -19,6 +19,8 @@ derive_builder = "0.12.0"
 fastly = "0.9.1"
 log = "0.4.17"
 log-fastly = "0.9.1"
+once_cell = { version = "1.18.0", features = [] }
```
jdno, line 22: Why is this syntax used? We're not enabling any optional features and afaik this does not disable default features. So why not just use `once_cell = "1.18.0"`?

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -127,6 +130,28 @@ fn rewrite_urls_with_plus_character(request: &mut Request) {
...
+/// Rewrite `/crates/{crate}/{version}/download` URLs to
+/// `/crates/{crate}/{crate}-{version}.crate`
+///
+/// cargo versions before 1.24 don't support placeholders in the `dl` field
+/// of the index, so we need to rewrite the download URL to point to the
+/// crate file instead.
+fn rewrite_download_urls(request: &mut Request) {
+    static RE: Lazy<Regex> = Lazy::new(|| {
+        Regex::new(r"^/crates/(?P<crate>[^/]+)/(?P<version>[^/]+)/download$").unwrap()
+    });
```
jdno, line 142: I'd prefer moving this into a constant, but also seem to vaguely remember that `once_cell` doesn't support that (yet)?

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -127,6 +130,28 @@ fn rewrite_urls_with_plus_character(request: &mut Request) {
...
+
+    let url = request.get_url_mut();
+    let path = url.path();
+
+    if let Some(captures) = RE.captures(path) {
+        let krate = captures.name("crate").unwrap().as_str();
```
jdno, line 148: Unwrapping here seems dangerous. I'm not sure how much debugging information we'll get if we just fail hard. Instead, I think it is better to change the return type of the function to `Result<(), Error>` and handle the error gracefully. And if we log an error with `error!`, we'll get to see that in the service logs.

jdno (changes requested):
Looks generally good, just the use of `unwrap` is something I'd like to fix or discuss.

terragrunt/modules/crates-io/compute-static/Cargo.toml
Turbo87: I honestly have no idea what happened here... 😂

I'll fix it :)

terragrunt/modules/crates-io/compute-static/src/main.rs
Turbo87: I don't think const is possible since the regex can't currently be constructed at buildtime

terragrunt/modules/crates-io/compute-static/src/main.rs
Turbo87: hmm, can you think of a scenario where the unwrap would panic? the outer if let ensures that the regex matches and both of the captures are non-optional so I think unwrap is safe here for this regex 🤔

terragrunt/modules/crates-io/compute-static/Cargo.toml
Turbo87: now that I think about it, I think I followed https://docs.rs/regex/latest/regex/#usage and used `cargo add` for the first time. is that the default for `cargo add`!? 🤔

terragrunt/modules/crates-io/compute-static/src/main.rs
jdno: Yeah that's what I thought. Then let's leave it like this for now. 👍

terragrunt/modules/crates-io/compute-static/src/main.rs
jdno: That's a good argument. I don't think this can panic in its current form, but I'm also somewhat paranoid about changes in the future and other people assuming, based on this example, that it's okay to `unwrap` in this binary.

jdno (commented):
I'm fine merging this as-is. Let me deploy it to staging so that we can smoketest the implementation before merging the pull request and deploying to production.

jdno:
This is now live in staging. 🚀

jdno:
Seems to work on staging, so I'll merge this and deploy to production.

jdno:
We fixed a last issue with the use of `let`, which is not supported by the used runtime.

Issue #372, opened 2023-12-11
Count Downloads Using CDN Logs


jdno:
## Problem

crates.io counts downloads by crate and version. This is currently done as part of the `/download` endpoint, which counts the download and then redirects the caller to the Content Delivery Networks (CDNs) for `static.crates.io`, from where the actual file is downloaded.

```mermaid
sequenceDiagram
	User->>crates.io: Requests crate
	crates.io->>crates.io: Counts crate and version
	crates.io->>User: Redirects user to static.crates.io
	User->>static.crates.io: Requests crate
	static.crates.io->>User: Serves crate file
```

Due to the volume of requests to the `/download` endpoint, counting the crate and its version in the application has a significant performance cost. Especially when traffic spikes, the application can struggle to keep up with requests, which in the worst case can cause a service outage.

### Key Objectives

1. Avoid hitting the web app for every crate download
2. Continue to count downloads by crate and version

### Desired Outcome

In the ideal scenario, we avoid hitting the web app for download requests altogether and go straight to the CDNs. We can achieve this by changing the `dl` field in the index's `config.json` to point to the CDN instead of the application. Full compatibility with existing behavior requires to rewrite some URLs, which [has already been implemented](https://github.com/rust-lang/simpleinfra/pull/365).

```mermaid
sequenceDiagram
	User->>static.crates.io: Requests crate
	static.crates.io->>User: Serves crate file
```

The CDNs could attempt to count download, but this is difficult because the CDNs are globally distributed. There is no single point that receives all the traffic, so download counts would need to be processed and merges somewhere else. That system would quickly face the same performance issues that `crates.io` currently faces.

We can use the request logs from the CDNs to count downloads in an asynchronous way. The CDNs produce a single log line per request. These logs are collected and uploaded periodically to a dedicated S3 bucket as a compressed archive.

Whenever a new archive is uploaded to the bucket, S3 can push an event into a SQS queue. `crates.io` can monitor the queue and pull incoming events. From the event, it can determine what files to fetch from S3, download and then parse them, and update the download counts in the database.

```mermaid
sequenceDiagram
	static.crates.io ->> S3: Uploads logs
	S3 ->> SQS: Queues event
	crates.io ->> SQS: Pulls event from queue
	crates.io ->> S3: Fetches new log file
	crates.io ->> crates.io: Parses log file
	crates.io ->> crates.io: Updates download counts
```

### Benefits

[...]

PR #374, opened 2023-12-19
Create AWS accounts for crates.io

jdno:
We are planning[^1] to count crate downloads using CDN logs. This requires new infrastructure, namely a SQS queue into which S3 can publish events and that crates.io can monitor.

[^1]: https://github.com/rust-lang/simpleinfra/issues/372

### DIFF
--- a/terragrunt/modules/aws-organization/groups.tf
+++ b/terragrunt/modules/aws-organization/groups.tf
@@ -28,6 +28,13 @@ resource "aws_identitystore_group" "billing" {
   description  = "People with access to the billing portal"
 }
 
+resource "aws_identitystore_group" "crates_io" {
+  identity_store_id = local.identity_store_id
+
+  display_name = "crates-io"
+  description  = "The crates.io team"
+}
+
 # The different permission sets a group may have assigned to it
 
 resource "aws_ssoadmin_permission_set" "administrator_access" {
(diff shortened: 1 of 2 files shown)


terragrunt/modules/aws-organization/groups.tf
```diff
@@ -112,6 +119,30 @@ locals {
...
+    # crates-io Production
+    {
+      account : aws_organizations_account.crates_io_prod,
+      groups : [
+        { group : aws_identitystore_group.infra-admins,
+        permissions : [aws_ssoadmin_permission_set.view_only_access, aws_ssoadmin_permission_set.administrator_access] },
+        { group : aws_identitystore_group.infra,
+        permissions : [aws_ssoadmin_permission_set.view_only_access] },
+        { group : aws_identitystore_group.crates_io,
+        permissions : [aws_ssoadmin_permission_set.view_only_access] },
```
Mark-Simulacrum, line 143: This seems like an excellent starting point, but I expect us to require expansion as needs arise - view only access is basically metadata only, not enough to do regular actions (e.g., re-drive messages on the eventual queue).

Mark-Simulacrum (approved):

terragrunt/modules/aws-organization/groups.tf
jdno: Agreed. I wasn't able to find prior art for this, and I don't think we actually discussed how exactly we want to deal with this. The separate accounts make it quite easy to grant pretty broad permissions, simplifying management. But I also feel like we need to make sure we have some oversight to prevent accidents or abuse. Let's discuss this further in #t-infra.

pietroalbini (approved):

PR #377, opened 2024-01-04
Initial SQS queue + user policy

Mark-Simulacrum:
This adds a new terragrunt module for the new crates.io account(s) and puts the SQS queue and related IAM policies into the module. However, it does not yet wire up the account or try to deploy this -- I haven't connected those bits yet and getting terragrunt to cooperate with me is usually pretty finicky. Hopefully this helps as a starting point though.

Remaining steps:

* Confirm whether the IAM user + hardcoded, non-rotating access key is the best way for Heroku crates.io to access this account (at least in the short term)
* Confirm this all actually works
* Create the resources

Possibly:

* Figure out if we want a dead letter queue. This would prevent a message crates.io doesn't know how to read from blocking all other messages in the queue (since it'll constantly fail and crates.io shouldn't call DeleteMessage on it). But, it's more complexity and it's not obvious that any particular message should be any different than others. I think it's not unreasonable to punt on setting this additional infrastructure up for now.
* Figure out alarming/metrics -- maybe crates.io should self-drive this with the queue metadata API, maybe we want to wire up datadog/grafana to read from this account.

IMO if we can, let's deploy this before we actually finish wiring up s3 -- crates.io can start reading from the queue and such early that way and kick the tires.

cc https://github.com/rust-lang/simpleinfra/issues/372

r? @jdno cc @Turbo87

### DIFF
--- a/terragrunt/modules/crates-io-logs/main.tf
+++ b/terragrunt/modules/crates-io-logs/main.tf
@@ -28,7 +28,7 @@ data "aws_iam_policy_document" "s3_push_to_queue" {
     condition {
       test     = "StringEquals"
       variable = "aws:SourceAccount"
-      values   = [data.aws_arn.account]
+      values   = [data.aws_arn.src_bucket.account]
     }
   }
 }
--- a/terragrunt/modules/crates-io-logs/main.tf
+++ b/terragrunt/modules/crates-io-logs/main.tf
@@ -1,0 +1,72 @@
+resource "aws_sqs_queue" "log_event_queue" {
+  name                      = "cdn-log-queue"
+  receive_wait_time_seconds = 20
+}
+
+resource "aws_sqs_queue_policy" "s3_push" {
+  queue_url = aws_sqs_queue.log_event_queue.id
+  policy    = data.aws_iam_policy_document.s3_push_to_queue.json
+}
+
+data "aws_iam_policy_document" "s3_push_to_queue" {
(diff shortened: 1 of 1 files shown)


terragrunt/modules/crates-io-logs/main.tf
```diff
@@ -0,0 +1,72 @@
...
+    resources = [aws_sqs_queue.log_event_queue.arn]
+    condition {
+      test     = "ArnLike"
+      variable = "aws:SourceArn"
+      values   = [data.aws_arn.src_bucket.arn]
+    }
+    condition {
+      test     = "StringEquals"
+      variable = "aws:SourceAccount"
+      values   = [data.aws_arn.account]
```
jdno, line 31:
```suggestion
      values   = [data.aws_arn.src_bucket.account]
```

jdno (changes requested):

jdno:
This looks like a good starting point. 👍  Can I help you with Terragrunt somehow?

Mark-Simulacrum:
Hm, I'm not sure there's a good intermediate step, I can probably copy what I did for bors with some adjustments for the deployed-ref business...

jdno (approved):
Let me just run with this for now and make a note to go over our documentation for Terragrunt.

jdno:
I'm gonna merge this and continue in a separate branch so that we review new changes more easily.

PR #387, opened 2024-02-06
Ensure package hash is stable

jdno:
Fastly added metadata collection[^1] to the CLI in version 10.6.0. The metadata includes the heap size during the build, which is not stable over time. Because of that, Terraform always detected a change to the compute package, even though no code had changed.

Disabling the metadata collection ensures that the package hash is stable between builds, thus reducing the noise in Terraform.

[^1]: https://developer.fastly.com/learning/tools/cli/#metadata-collection


jdno:
I've reported this as an issue here: https://github.com/fastly/cli/issues/1125.

Mark-Simulacrum (approved):

PR #389, opened 2024-02-16
Clean up unused checkouts on dev-desktops

jdno:
User might checkout repositories on the dev-desktops, work on them for a while, and then move on to other projects. The unused checkouts, in particular their `build` or `target` directories, can use up a lot of space that we want to reclaim periodically. A script has been created that finds all cache directories, checks recursively if any file in their root directory has been modified within a given time frame, and if not deletes the cache directory.


ansible/roles/dev-desktop/files/clean-unused-checkouts.sh
```diff
@@ -0,0 +1,70 @@
...
+
+  # If no files have been modified in the last 90 days, then the project is
+  # considered old.
+  if [[ -z "${modified}" ]]; then
+    echo "${directory}"
+  fi
+done)
+
+# Delete the build directories in the unused checkouts
+for directory in $unused_cache_directories; do
```
Mark-Simulacrum, line 63: Is there a "$unused[*]" or something needed here?

(Maybe we should write this script in Rust, like the ssh auth scripts... If this runs as root this seems easy to get wrong and cause us, at least theoretical, trouble).

jdno: > Is there a "$unused[*]" or something needed here?
>
> (Maybe we should write this script in Rust, like the ssh auth scripts... If this runs as root this seems easy to get wrong and cause us, at least theoretical, trouble).

No, since `$unused_cache_directories` is a multiline string and not a Bash array. 😅 But that question and specific implementation detail is probably not a bad reason to rewrite this in Rust. Would also make it much easier to write tests for the functionality. I'll give this a try and see how it would look...

Mark-Simulacrum (approved):

ansible/roles/dev-desktop/templates/clean-unused-checkouts.sh
```diff
@@ -0,0 +1,74 @@
...
+done
+
+# Find all build or target directories created by users
+#
+# This command combines (`-o`) two different conditions to find all build and
+# target directories that users have created. Within each home directory, we
+# recursively look for directories that either have a file named `x.py` and a
+# directory named `build`, or a file named `Cargo.toml` and a directory named
+# `target`.
+all_cache_directories=$(find /home -type d \( -name build -execdir test -f "x.py" \; -o -name target -execdir test -f "Cargo.toml" \; \) -print | sort | uniq)
```
compiler-errors, line 49: You may also want to sweep `build-rust-analyzer` too, which is a secondary build dir recommended for r-a workflows: https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc

spastorino:
Great @jdno. I was wondering if on top of removing old builds, we shouldn't have certain amount of free space always ensured just in case some of us are creating new build and may run out of space.
Maybe the script can check for certain amount of free disk space threshold and if we don't have enough consider removing from the oldest to the newest until we can free up to that threshold.

ansible/roles/dev-desktop/templates/clean-unused-checkouts.sh
jdno: I'll make a note of this in #390, which tracks rewriting the script in Rust. I fear adding more conditions to the `find` command will get unmanageable quickly. 🙈

jdno:
> Great @jdno. I was wondering if on top of removing old builds, we shouldn't have certain amount of free space always ensured just in case some of us are creating new build and may run out of space.

There are two separate issues that we want to handle differently, I think. First, we don't want to run out of disk space on the root partition `/` to avoid locking up the machine. We could do this by extending this script, for example, but I think ultimately we want to move the `/` and `/home` partitions to different disks. Second, we want people to always be able to create a new checkout (within their quota). Cleaning up "unused" projects more aggressively might help. 🤔

PR #393, opened 2024-03-05
Increase robustness of SSL renewing

shepmaster:
The playground had an error renewing:

```
Error -> One or more domains had a problem:
error: 0 :: POST :: https://acme-v02.api.letsencrypt.org/acme/finalize/514094497/249375107307 :: urn:ietf:params:acme:error:rateLimited :: Service busy; retry later., url:
```

Our renew timer runs once a week at midnight UTC, and my guess is that lots of other servers across the world are doing the same thing causing temporary overloads. These changes are intended to help spread out the load a bit and retry when an error happens anyway.


jdno (commented):
This is part of #311 as well, isn't it?

shepmaster:
Yep. This should be merged first and then I can rebase #311 .

jdno (approved):
Okay then let's do that. 😁

PR #397, opened 2024-03-14
Remove the `cancel-outdated-builds` action

Kobzol:
This action was only [used](https://github.com/search?q=org%3Arust-lang%20cancel-outdated-builds&type=code) in Clippy, where I have removed its usage in https://github.com/rust-lang/rust-clippy/pull/12477#event-12116780479.

I think that we can remove it now, since it's supported natively by Github Actions using the `concurrency` group.


jdno (approved):
The action is [still used](https://github.com/search?q=path%3A%2F%5E.github%5C%2F%2F+%22uses%3A+rust-lang%2Fsimpleinfra%2Fgithub-actions%2Fcancel-outdated-builds%22&type=code) in a lot of forks. Sadly all locked to `@master`. 🙈

But given that we don't seem to be using it within the `rust-lang` organization, I think it should be safe to remove.

PR #398, opened 2024-03-15
Grant access to global SSM parameters

jdno:
When deploying the playground, access to the global SSM parameters in the `/{staging,prod}/ansible/all/*` namespace is required.


pietroalbini (approved):
bors merge

PR #402, opened 2024-04-04
Update cargo resolver version

kapare:
Calling cargo command into the project was giving this warning:

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2021 which implies `resolver = "2"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
note: for more details see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
```

The workspace.resolver is updated to version 2 and each crates set edition = 2021


jdno:
Thanks! 🙂

PR #404, opened 2024-04-10
monitoring: Remove obsolete "cratesio/SlowDownloads" alerting rule

Turbo87:
The crate downloads are no longer going through the crates.io API. Since this alerting rule is based on the crates.io API traffic, it no longer serves its purpose. Whether we can implement something similar based on the CDN traffic logs would probably have to be figured out by the infra team 😅

### DIFF
--- a/ansible/playbooks/monitoring.yml
+++ b/ansible/playbooks/monitoring.yml
@@ -189,12 +189,6 @@
               labels:
                 dispatch: pagerduty-cratesio
 
-            - alert: SlowDownloads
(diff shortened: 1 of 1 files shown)


jdno (approved):
We can remove the rule. :+1: But if we need to recreate it based on the logs, we should create a tracking issue. Otherwise we'll forgot.

Turbo87:
I expect us to revisit the whole alerting situation once DataDog is fully operational with the whole team having access. I guess the "slow" downloads was primarily relevant for the previous situation where such an alert was actionable. We will have to see what alerts will make sense in the future.

jdno:
I ran Ansible and the alert has been removed from Prometheus. :shipit:

Issue #405, opened 2024-04-13
Log user-agents for Fastly CDN logs


Mark-Simulacrum:
Currently our CloudFront logs contain user agents, but the Fastly logs do not. We should fix that, probably by editing something around here: https://github.com/rust-lang/simpleinfra/blob/86a318bdd2ca6092a6bcc3c4f81ef541e7bd78b7/terragrunt/modules/release-distribution/fastly-static.tf#L108 to include a user agent (https://www.fastly.com/documentation/reference/api/logging/s3/)

PR #406, opened 2024-04-16
Stream request logs for static.crates.io from Fastly to Datadog

jdno:
crates.io recently [implement a few changes](https://blog.rust-lang.org/2024/03/11/crates-io-download-changes.html) that make `cargo` download crates directly from `static.crates.io` without hitting the API of crates.io first. This increases the performance and scalability of crates.io, but means that requests are no longer logged by the application. We want to restore the previous behavior by aggregating request logs from our CDNs in Datadog.

The Fastly service has been extended to include more information in its request logs and to stream them to Datadog. We're also tagging the logs with Datadog's [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging) to provide visibility across the app (`crates.io`), service (`static.crates.io`), and different environments.

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -48,8 +48,8 @@ fn main(request: Request) -> Result<Response, Error> {
 fn init_logging(config: &Config) {
     Logger::builder()
         .max_level(LevelFilter::Debug)
-        .endpoint(config.request_logs_endpoint.clone())
-        .default_endpoint(config.service_logs_endpoint.clone())
+        .endpoint(config.s3_request_logs_endpoint.clone())
+        .default_endpoint(config.s3_service_logs_endpoint.clone())
         .echo_stdout(true)
         .init();
 }
@@ -218,7 +218,15 @@ fn build_and_send_log(log_line: LogLineV1Builder, config: &Config) {
     match log_line.build() {
         Ok(log) => {
             let versioned_log = LogLine::V1(log);
-            info!(target: &config.request_logs_endpoint, "{}", json!(versioned_log).to_string())
+
+            [
+                &config.datadog_request_logs_endpoint,
+                &config.s3_request_logs_endpoint,
+            ]
+            .iter()
+            .for_each(|endpoint| {
+                info!(target: endpoint, "{}", json!(versioned_log).to_string());
+            });
         }
         Err(error) => {
             warn!("failed to serialize request log: {error}");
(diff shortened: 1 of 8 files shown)


terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -1,18 +1,22 @@
...
 
 use crate::config::Config;
-use crate::log_line::{LogLine, LogLineV1Builder};
+use crate::log_line::{HttpDetailsBuilder, LogLine, LogLineV1Builder, TlsDetailsBuilder};
 
 mod config;
 mod log_line;
 
+const DATADOG_APP: &str = "crates.io";
+const DATADOG_SERVICE: &str = "static.crates.io";
```
jdno, line 18: Originally, these two static constants were defined in Terragrunt. But they don't change depending on the environment and hardcoding them removes a potential panic if they'd accidentally got deleted.

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -48,20 +52,52 @@ fn main(request: Request) -> Result<Response, Error> {
...
 /// Collect data for the logs from the request
-fn collect_request(request: &Request) -> LogLineV1Builder {
-    LogLineV1Builder::default()
+fn collect_request(config: &Config, request: &Request) -> LogLineV1Builder {
+    let http_details = HttpDetailsBuilder::default()
+        .protocol(http_version_to_string(request.get_version()))
+        .referer(
+            request
+                .get_header("Referer")
+                .and_then(|s| s.to_str().ok())
```
jdno, line 69: `.ok()` is used here to prevent issues with the logs causing a request to fail.

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -48,20 +52,52 @@ fn main(request: Request) -> Result<Response, Error> {
...
+        .build()
+        .ok();
+
+    let log_line = LogLineV1Builder::default()
+        .ddtags(format!("app:{},env:{}", DATADOG_APP, config.datadog_env))
+        .service(DATADOG_SERVICE)
         .date_time(OffsetDateTime::now_utc())
-        .url(request.get_url_str().into())
+        .edge_location(var("FASTLY_POP").ok())
+        .host(request.get_url().host().map(|s| s.to_string()))
```
Turbo87, line 92: is there a reason for including `host` if it's essentially derived from `url`?

Mark-Simulacrum:
I guess this is a direct stream, right? So this doesn't address https://github.com/rust-lang/simpleinfra/issues/405?

jdno:
The same log format is used for the stream and the files that we write to S3. So after this change, those will include the `User-Agent` header as well.

jdno:
Does it make sense to bump the version of the log format to `version: 2` with these changes?

Turbo87:
> Does it make sense to bump the version of the log format to `version: 2` with these changes?

that would require code updates and us deploying those updates to crates.io before the log format can be updated. since the changes are additive and non-breaking I don't think we need a new version.

PR #411, opened 2024-04-27
Grant ehuss and spastorino access to triagebot logs

Mark-Simulacrum:
This grants scoped-down access to just the triagebot log group in our legacy AWS account. The console appears to still be usable with the access granted here, so it seems good enough. Note that we do grant ability to list all log groups but this is not particularly meaningful.

cc @ehuss @spastorino

### DIFF
--- a/terragrunt/accounts/root/aws-organization/terragrunt.hcl
+++ b/terragrunt/accounts/root/aws-organization/terragrunt.hcl
@@ -25,7 +25,7 @@ inputs = {
       given_name  = "Mark",
       family_name = "Rousskov"
       email       = "mark.simulacrum@gmail.com"
-      groups      = ["infra", "infra-admins"]
+      groups      = ["infra", "infra-admins", "triagebot"]
     }
     "rylev" = {
       given_name  = "Ryan",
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
Although now that I look I think I mis-remembered that @spastorino was doing a bunch of work with triagebot recently... let me actually move that to @apiraino who actually asked for this access.

spastorino:
> Although now that I look I think I mis-remembered that @spastorino was doing a bunch of work with triagebot recently... let me actually move that to @apiraino who actually asked for this access.

For what is worth, I'm not doing any triagebot work since a long while and I'm not really sure what is this access about.

apiraino:
> For what is worth, I'm not doing any triagebot work since a long while and I'm not really sure what is this access about.

@spastorino this is a limited read-only access to the triagebot AWS logs. If you don't work on the triagebot, you don't need this, I think. You're also in "alumni" status in the triagebot team.

@Mark-Simulacrum ok if I remove him (i.e. submitting an update to this patch)?

ehuss:
The update should already have been done in https://github.com/rust-lang/simpleinfra/pull/412.

PR #413, opened 2024-05-02
Create bucket to store downloads statistics

jdno:
The crates.io team is considering adding a long-term archive for download statistics so that historic data can be removed from the database. The proposal in rust-lang/crates.io#3479 is to put daily CSV files into an S3 bucket.

A new bucket has been created for this purpose. Other than the existing buckets, this one is not publicly accessible. The bucket has been created in the new `crates-io-staging` account so that the team can more easily access it.


jdno:
I've opened this as a draft PR so that the configuration can be reviewed while @Turbo87 is testing the implementation.

Mark-Simulacrum (commented):
If we're creating a new bucket, can we not put it in us-west-1? Especially for long-term archival that feels like it's mostly just paying extra for no good reason (crates.io runs in us-east-1 anyway, AFAIK).

jdno:
Creating the bucket in a different region than the rest of the resources requires changes to our Terragrunt setup. Once #414 is merged, we can rebase and move the bucket.

Mark-Simulacrum:
Moving buckets isn't possible (at least without talking to support?) but I guess we can create a new one and drop this one. Approved 414.

jdno:
Sorry, that's what I meant. Since it's staging and we're still setting everything up, "move" meant "destroy and recreate". 😅

jdno:
Rebased the PR and recreated the bucket in `us-east-1`.

Turbo87 (approved):

PR #417, opened 2024-05-14
Increase disk size of dev-desktops

jdno:
The dev-desktops were running out of disk space, particularly the `eu-2` instance. We've doubled the size of the disks to 2TB, which will cost us around $300-400 more per months. This should be enough for a while and give us room to onboard more users, for example participants of Google Summer of Code.


jdno:
I've documented the process here: https://github.com/rust-lang/infra-team/blob/master/service-catalog/dev-desktops/how-to-increase-disk-size.md

PR #421, opened 2024-06-01
crates-io: Add support for `db-dump.zip` files

Turbo87:
see https://github.com/rust-lang/crates.io/pull/8748

### DIFF
--- a/terragrunt/modules/crates-io/s3-static.tf
+++ b/terragrunt/modules/crates-io/s3-static.tf
@@ -13,13 +13,24 @@ resource "aws_s3_bucket" "static" {
     max_age_seconds = 3000
   }
 
-  // Keep only the live db-dump.tar.gz and the previous day's version, removing
+  // Keep only the live db-dumps and the previous day's versions, removing
   // all the other ones. This is needed because we don't want this file to be
   // versioned, while all the other ones in the bucket should be versioned.
   lifecycle_rule {
     id      = "purge-db-dump"
     enabled = true
     prefix  = "db-dump.tar.gz"
+
+    abort_incomplete_multipart_upload_days = 1
+    noncurrent_version_expiration {
+      days = 1
+    }
+  }
+
+  lifecycle_rule {
+    id      = "purge-db-dump-zip"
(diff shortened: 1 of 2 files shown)


terragrunt/modules/crates-io/s3-static.tf
```diff
@@ -27,6 +27,17 @@ resource "aws_s3_bucket" "static" {
...
+  lifecycle_rule {
+    id      = "purge-db-dump-zip"
+    enabled = true
+    prefix  = "db-dump.zip"
+
+    abort_incomplete_multipart_upload_days = 1
+    noncurrent_version_expiration {
+      days = 1
+    }
+  }
```
jdno, line 39: I'd like to avoid the duplication here. Can we use a [dynamic block](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) for this instead?

```hcl
dynamic "lifecycle_rule" {
  for_each = toset(["db-dump.tar.gz", "db-dump.zip"])

  content {
    id      = "purge-${replace(lifecycle_rule.key, ".", "-")}"
    enabled = true
    prefix  = lifecycle_rule.key

    abort_incomplete_multipart_upload_days = 1
    noncurrent_version_expiration {
      days = 1
    }
  }
}
```

jdno (changes requested):

terragrunt/modules/crates-io/s3-static.tf
Turbo87: > Can we use a [dynamic block](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) for this instead?

you tell me, I have no idea how to check if this works or not 😂

jdno: Sorry, let me rephrase the question. Can you please change it to a dynamic block? 😂 I already tested that the code above works.

Turbo87: done :)

jdno (approved):
This is live on staging. :shipit:

PR #422, opened 2024-06-04
Manage win.rustup.rs in Terragrunt

jdno:
`win.rustup.rs` is an old CloudFront distribution that is not yet managed in Terraform or Terragrunt. A new Terragrunt module has been created so that we can easily create a new staging environment.

The module copies the existing functionality, but uses a response header policy instead of a Lambda function to set the `Content-Disposition` header.

This is currently blocking https://github.com/rust-lang/rustup/pull/3840.


terragrunt/modules/win-rustup-rs/lambda.tf
```diff
@@ -0,0 +1,13 @@
...
+  source = "../aws-lambda"
+
+  providers = {
+    aws = aws.us-east-1
+  }
+
+  name       = "${local.human_readable_name}--origin-request"
+  source_dir = "lambdas/origin-request"
+  handler    = "index.handler"
+  runtime    = "nodejs16.x"
```
pietroalbini, line 11: This runtime is going to be deprecated in a week (https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html), maybe we can switch to a new one?

terragrunt/modules/win-rustup-rs/main.tf
```diff
@@ -0,0 +1,71 @@
...
+      cookies {
+        forward = "none"
+      }
+    }
+
+    lambda_function_association {
+      event_type   = "origin-request"
+      lambda_arn   = module.origin_request.version_arn
+      include_body = false
+    }
```
pietroalbini, line 57: Doing this is not a hard requirement, but given that the function is so simple I'd consider switching to a CloudFront Function rather than a Lambda@Edge. The main reason is that it's easier to deploy (just add a `aws_cloudfront_function` rather than dealing with the whole lambda module), but it's also cheaper and faster.

pietroalbini (approved):

PR #428, opened 2024-06-18
Create S3 bucket for Rustup build artifacts

jdno:
A new S3 bucket is being created as part of the effort to rebuild the Rustup release process. The bucket will store the build artifacts for Rustup, which the GitHub Actions in rust-lang/rustup will produce.

Since the bucket is tied to a GitHub repository, only a single bucket in the production environment is being created.

### DIFF
--- a/terragrunt/modules/rustup/s3.tf
+++ b/terragrunt/modules/rustup/s3.tf
@@ -1,0 +1,32 @@
+# The rust-lang/rustup repository on GitHub builds and uploads Rustup artifacts
+# to this S3 bucket.
+resource "aws_s3_bucket" "builds" {
+  provider = aws.us-east-1
+
+  bucket = "rustup-builds"
+}
+
+module "aws_iam_user" {
+  source = "../gha-iam-user"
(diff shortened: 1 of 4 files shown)


jdno:
I've created this as a draft PR, since applying it will overwrite the secrets in `rust-lang/rustup`.

terragrunt/modules/rustup/s3.tf
```diff
@@ -0,0 +1,51 @@
+# The rust-lang/rustup repository on GitHub builds and uploads Rustup artifacts
+# to this S3 bucket.
+resource "aws_s3_bucket" "builds" {
+  provider = aws.us-east-1
+
+  bucket = "rustup-builds"
+}
+
+module "aws_iam_user" {
+  source = "../gha-iam-user"
```
Mark-Simulacrum, line 10: We should really be using OIDC for anything new, not IAM users.

Mark-Simulacrum (approved):

Mark-Simulacrum (approved):

PR #438, opened 2024-07-06
CI: Add `terraform validate` command

Turbo87:
see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/Proposal.3A.20AWS.20SNS.20for.20Crate.20Actions.3F/near/449467635 😉

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: [ push, pull_request ]
 
 jobs:
   rust:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -25,6 +25,9 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
+      - name: Ensure the terraform files are semantically valid
+        run: terraform validate
```
Nilstrieb, line 29: It doesn't do anything here at the top-level, it needs to be run in a Terraform configuration directory.

Turbo87:
Nevermind... looks like this will have to be run in each directory that contains `.tf` files... 😩

PR #439, opened 2024-07-06
crates-io: Disable versioning for `rss/` contents

Turbo87:
Note that I have absolutely no idea if the content of this PR works or not since there does not seem to be a way for me to check this, but the intention is to disable S3 versioning for all files inside the `rss/` folder on S3.

r? @rust-lang/infra

### DIFF
--- a/terragrunt/accounts/legacy/crates-io-staging/crates-io/.terraform.lock.hcl
+++ b/terragrunt/accounts/legacy/crates-io-staging/crates-io/.terraform.lock.hcl
@@ -5,6 +5,7 @@ provider "registry.terraform.io/fastly/fastly" {
   version     = "5.0.0"
   constraints = "5.0.0"
   hashes = [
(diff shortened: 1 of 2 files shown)


pietroalbini:
You cannot disable versioning for just a subset of objects in a bucket, versioning is a global setting that applies to the whole bucket. Pushed a commit with a working version of this configuration, which deletes non-current versions of the feed after one day.

I'll let someone else in t-infra apply this (I don't have fastly keys on the machine I'm currently using).

jdno:
The changes here are live on staging.

MarcoIeni:
> The changes here are live on staging.

Should we merge this and apply to prod as well?

Right now `terragrunt plan` is not clean on master because it detects this change.

MarcoIeni:
I'm deploying to prod, then 🎉

MarcoIeni:
Done 👍

PR #440, opened 2024-07-06
Enable public read access to the release logs

pietroalbini:
r? @Mark-Simulacrum 

I didn't apply this yet.


Mark-Simulacrum (approved):
I think we should probably spot check a release log to check for any secret leakage, especially given sensitivity.. but I don't recall any problems so I think this is good to go.

pietroalbini:
Given we're going the route of invoking the release process through a local script authenticating with IAM I don't think there's a need for this anymore.

PR #443, opened 2024-07-14
Grant permissions to the release team

pietroalbini:
This doesn't grant access to members of t-release yet, just sets up the permission set and the group.

### DIFF
--- a/terragrunt/modules/aws-organization/groups.tf
+++ b/terragrunt/modules/aws-organization/groups.tf
@@ -42,6 +42,13 @@ resource "aws_identitystore_group" "triagebot" {
   description  = "The triagebot maintainers"
 }
 
+resource "aws_identitystore_group" "release" {
+  identity_store_id = local.identity_store_id
+
+  display_name = "release"
+  description  = "The release team"
+}
+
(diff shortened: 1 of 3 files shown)


terragrunt/modules/aws-organization/groups.tf
```diff
@@ -156,6 +163,62 @@ resource "aws_ssoadmin_permission_set_inline_policy" "triagebot_access" {
...
+    Statement = [
+      {
+        Effect   = "Allow"
+        Action   = "lambda:InvokeFunction"
+        Resource = "arn:aws:lambda:us-west-1:890664054962:function:start-release"
+      },
+      {
+        Effect = "Allow"
+        Action = [
+          // This must NOT include codebuild:StartBuild, starting builds must
```
Mark-Simulacrum, line 188: Should we explicitly deny it perhaps?

Mark-Simulacrum (approved):

terragrunt/modules/aws-organization/groups.tf
pietroalbini: Done. Also removed a permission that is not actually needed.

PR #447, opened 2024-07-18
give marco ieni terraform access

MarcoIeni:


terraform/shared/terraform-state.tf
```diff
@@ -10,6 +10,7 @@ locals {
     "pietroalbini",
     "simulacrum",
     "jdn",
+    "marcoieni",
```
MarcoIeni, line 13: I'm not sure if we want to add this now or in a later pr

terraform/team-members-access/_users.tf
```diff
@@ -7,6 +7,7 @@ locals {
     "pietroalbini"  = [aws_iam_group.infra_admins.name],
     "simulacrum"    = [aws_iam_group.infra_admins.name],
     "jdn"           = [aws_iam_group.infra_admins.name],
+    "marcoieni"     = [aws_iam_group.infra_admins.name],
```
MarcoIeni, line 10: this should create a new aws account I guess. not sure if it's enough.

terragrunt/accounts/root/aws-organization/terragrunt.hcl
```diff
@@ -81,5 +81,11 @@ inputs = {
       email = "apiraino@protonmail.com"
       groups = ["triagebot"]
     }
+    "marcoieni" = {
+      given_name  = "Marco",
+      family_name = "Ieni"
+      email       = "marcoieni@rustfoundation.org"
+      groups      = ["infra", "infra-admins"]
+    }
```
MarcoIeni, line 89: not sure if this is required.

jdno (approved):

PR #449, opened 2024-07-18
fix(aws-cred): accept any mfa name

MarcoIeni:
MFA can have different names. JD said this is a recent change.
With this fix, this script works for me 👍

### DIFF
--- a/aws-creds.py
+++ b/aws-creds.py
@@ -72,10 +72,7 @@ def load_mfa_device():
 
     serial = None
     for device in mfa_out["MFADevices"]:
-        # Only TOTP devices are supported
         user = device["UserName"]
-        if not device["SerialNumber"].endswith(f":mfa/{user}"):
-            continue
 
         # Only a single TOTP device is supported
         if serial is None:


aws-creds.py
```diff
@@ -74,8 +74,6 @@ def load_mfa_device():
     for device in mfa_out["MFADevices"]:
         # Only TOTP devices are supported
```
Mark-Simulacrum, line 75: Seems like this comment is probably stale? But it's not clear to me -- I guess we'd now accept e.g. `u2f` devices too. But those probably won't work with this script, since we will later request a TOTP code specifically?

MarcoIeni: I'm not sure because I didn't test this.
We had a look at this script together with JD during my onboarding and we noticed there are many comments that aren't valid anymore. For example, now users can have more than one MFA device in AWS.

In this PR I just deleted the lines that block me from logging in, but certainly it would be nice to read this script entirely and re-verify all the assumptions 👍
Also, in the future we might want to use sso so maybe most of this script becomes not necessary? 🤔

Mark-Simulacrum: I'd like to see this particular comment dropped at least, but otherwise happy to merge and see if that breaks anything.

jdno (approved):

PR #455, opened 2024-07-26
fix(docs-rs-staging): fix terragrunt configuration

MarcoIeni:


terragrunt/accounts/docs-rs-staging/account.json
```diff
@@ -0,0 +1,11 @@
+{
+    "aws": {
+        "profile": "docs-rs-staging",
+        "regions": [
+            {
+                "region": "us-east-1",
+                "alias": "us-east-1"
```
jdno, line 7:
```suggestion
```

The default region must not have an alias, otherwise authentication will not work. If you add the alias here, we end up with a automatically generated `aws` provider without any configuration and a `aws.us-east-1` provider with the correct configuration. But no resource is configured to use this provider, they all default to the `aws` provider.

MarcoIeni: Thanks! What do you think about validating this in `terragrunt-locals.py`?

Issue #436, opened 2024-07-04
Set up CDN for crates.io's download statistics


jdno:
The infra-team recently [created an S3 bucket](https://github.com/rust-lang/simpleinfra/pull/413) for crates.io so that it can export its daily downloads statistics from the database. The crates.io team wants to make these statistics publicly available, which requires us serve them through a CDN.

Two different options [have been discussed](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/archiving.20.60version_downloads.60.20to.20S3/near/440480148):

1. Add the new bucket to `static.crates.io` as a new origin
2. Create a new CDN just for the statistics

No conclusion has been reached and resolving this discussion is one of the first tasks of this issue.

## Tasks

- [x] Explore both options and propose a solution to the infra-team
- [ ] Set up CDN in staging environment
- [ ] Confirm working functionality with crates.io team
- [ ] Set up CDN in production environment

MarcoIeni:
I'll work on it!

MarcoIeni:
In the following I will list pros (👍) and cons (👎) of each approach.
I'm writing them down to:
- confirm I understood the issue
- take an informed decision

> [!NOTE]
> Everyone, please bare in mind that I started this week and I'm not familiar with the current setup of crates.io infrastructure, so I might be missing some important details and I might be wrong in some of the points I'm making.
> So correct me if I'm wrong please 🙏

## Approach 1: Creating a new CloudFront distribution

Serve this new bucket as something like `downloads.crates.io`.
Proposed [here](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/archiving.20.60version_downloads.60.20to.20S3/near/440490177).

### 👍 Easier way to track the traffic. 

We can easily track the traffic of the new distribution (we need to start tracking `downloads.crates.io`). So that we know that the traffic of `static.crates.io` only contains traffic of crates (not downloads).
This means there's no need to update the current dashboards that track the traffic of `static.crates.io`.

> [!NOTE]
> I might be totally wrong about this because maybe it's easier to reuse the same mechanism we have in place for `static.crates.io`.
> Maybe start tracking a new distribution takes a lot of work or resources.

### 👍 Reuse the same code

As JD said:

> right now we have a 1:1 mapping between buckets and distributions.

So this approach is consistent with the current setup because it doesn't break the 1:1 mapping we have for the other buckets.

[...]

PR #469, opened 2024-08-02
feat: serve version downloads `index.html`

MarcoIeni:
- [x] test this after applying it. both in cloudfront and fastly

blocked by https://github.com/rust-lang/crates.io/pull/9207

related to https://github.com/rust-lang/simpleinfra/issues/436

### DIFF
--- a/terragrunt/modules/crates-io/cloudfront-functions/static-router.js
+++ b/terragrunt/modules/crates-io/cloudfront-functions/static-router.js
@@ -1,10 +1,16 @@
 function handler(event) {
     var request = event.request;
+    var versionDownloads ='/archive/version-downloads/';
 
     // URL-encode the `+` character in the request URI
     // See more: https://github.com/rust-lang/crates.io/issues/4891
     if (request.uri.includes("+")) {
         request.uri = request.uri.replace("+", "%2B");
+    } else if (request.uri === versionDownloads) {
+        request.uri += 'index.html';
+        return request;
+    } else if (request.uri === '/archive/version-downloads') {
+        return permanentRedirect(versionDownloads);
     }
 
     // cargo versions before 1.24 don't support placeholders in the `dl` field
@@ -18,4 +24,14 @@ function handler(event) {
     }
 
     return request;
+}
+
+function permanentRedirect(destination) {
+    return {
+        statusCode: 301,
+        statusDescription: 'Moved Permanently',
+        headers: {
+            'location': { value: destination },
+        },
+    };
 }
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -1,4 +1,5 @@
-use fastly::http::{Method, StatusCode, Version};
+use fastly::convert::ToHeaderValue;
+use fastly::http::{header, Method, StatusCode, Version};
 use fastly::{Error, Request, Response};
 use log::{info, warn, LevelFilter};
 use log_fastly::Logger;
@@ -16,6 +17,7 @@ mod log_line;
 
 const DATADOG_APP: &str = "crates.io";
 const DATADOG_SERVICE: &str = "static.crates.io";
+const VERSION_DOWNLOADS: &str = "/archive/version-downloads/";
 
 #[fastly::main]
 fn main(request: Request) -> Result<Response, Error> {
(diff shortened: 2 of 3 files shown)


terragrunt/modules/crates-io/cloudfront-functions/static-router.js
```diff
@@ -5,6 +5,9 @@ function handler(event) {
     // See more: https://github.com/rust-lang/crates.io/issues/4891
     if (request.uri.includes("+")) {
         request.uri = request.uri.replace("+", "%2B");
+    } else if (request.uri === '/archive/version-downloads/') {
```
Turbo87, line 8: is this guaranteed to always end with a slash?

terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -168,6 +169,19 @@ fn rewrite_urls_with_plus_character(request: &mut Request) {
...
 }
 
+/// Rewrite `/archive/version-downloads/` URLs to `/archive/version-downloads/index.html`
+///
+/// In this way, users can see what files are available for download.
+fn rewrite_version_downloads_urls(request: &mut Request) {
+    let url = request.get_url_mut();
+    let path = url.path();
+
+    if path == "/archive/version-downloads/" {
```
Turbo87, line 179: same question here, I guess :)

terragrunt/modules/crates-io/cloudfront-functions/static-router.js
jdno: Depends on the user and what they type into the address bar. 😅

I think the question is more: Do we want to support the URL without a trailing slash as well? Personally, I'd say No until someone complains. While it's not _that_ more work to change this into a regex, a simple comparison is easier to reason about and anything else might be premature optimization...

Turbo87: I'd prefer to support both. It's too confusing otherwise and the point of the index file is discoverability.

Turbo87: FWIW I'd also be fine with a redirect from no-slash to slash :)

LawnGnome: I'd prefer to redirect to include the slash so we can use relative links in the generated HTML, if that's not too hard.

MarcoIeni: Isn't it what the code does at the moment?
Or am I missing something?

Turbo87: > Isn't it what the code does at the moment?

it currently serves the HTML file at both locations.

- at `/archive/version-downloads/` a link like `<a href="2024-08-20.csv">` would lead to `/archive/version-downloads/2024-08-20.csv`
- at `/archive/version-downloads` a link like `<a href="2024-08-20.csv">` would lead to `/archive/2024-08-20.csv`

that's why Adam prefers an HTTP redirect ([301 status code + `Location` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301#server_response)) for the non-slash location, instead of serving the HTML there directly.

terragrunt/modules/crates-io/cloudfront-functions/static-router.js
```diff
@@ -19,3 +25,13 @@ function handler(event) {
...
+
+function permanentRedirect(destination) {
+    return {
+        statusCode: 301,
+        statusDescription: 'Moved Permanently',
+        headers: {
+            'location': { value: destination },
+        },
+    };
+}
```
MarcoIeni, line 37: found the syntax here: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/functions-tutorial.html

terragrunt/modules/crates-io/cloudfront-functions/static-router.js
MarcoIeni: Makes sense.
Ready for another look 👍

Turbo87 (approved):
LGTM

I guess we can try this out on staging first :)

MarcoIeni:
tested in staging 👍 
https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/archiving.20.60version_downloads.60.20to.20S3/near/463723272

MarcoIeni:
Deployed to prod as well!

PR #474, opened 2024-08-08
chore(docs-rs-staging): delete unused db

MarcoIeni:
Discussed [here](https://rust-lang.zulipchat.com/#narrow/stream/356853-t-docs-rs/topic/Can.20we.20delete.20.60docs-rs-prod.60.20db.3F).
I stopped the DB and nobody complained 👍


MarcoIeni:
The final snapshot is called `docs-rs-final-snapshot` if anybody is interested in the data.

PR #480, opened 2024-08-14
Add earlyoom to dev desktop

Noratrieb:
Sometimes the dev desktop just OOMs and dies because someone committed a happy little accident, as one does. Luckily Linux has an OOM killer, sadly it's pretty useless in practice.

Luckily there's a userspace workaround! It works really well.

I don't know if we want any config for it. I set the freeSwapThreshold to 100 on my own computer to never swap before killing, I don't know if that would matter for the dev desktop. I think it makes sense to keep it at the default for now unless that causes problems.


MarcoIeni:
Hey Nora, thank you for contributing!
I tested it on our staging dev desktops by running `./x check` and earlyoom killed the process almost immediately (the staging desktop only has 1GB of memory 😅).

So this looks good to me. Let's try it with default settings as you said and let's see how it goes. If people complain that their processes get killed too often we can re-evaluate this :)

Just one thing. Instead of adding a task like you did, I think it's cleaner if you add `earlyoom` to this list:
https://github.com/rust-lang/simpleinfra/blob/5bbba89f181fc801116de915b0beae90e436d7db/ansible/roles/dev-desktop/tasks/dependencies.yml#L11

What do you think?

PR #488, opened 2024-08-22
chore: update discord-mods-bot to aws 4.67

MarcoIeni:
plan is clean. merging


terraform/discord-mods-bot/_terraform.tf
```diff
@@ -29,6 +29,5 @@ data "terraform_remote_state" "shared" {
 }
 
 provider "aws" {
-  profile = "default"
```
MarcoIeni, line 32: I removed profile because it creates issues for me. If you think we should have it and I'm doing something wrong (probably it's the case 😅), let me know 👍

PR #516, opened 2024-08-26
add Marco to release_keys_bucket_users

marcoieni:
I can't apply this, so whoever approves it, please apply as well and merge 👍


jdno (approved):
I applied the change. :shipit:

PR #540, opened 2024-08-28
chore(release-distribution): update `doc-router` to node 20

MarcoIeni:


MarcoIeni:
Tested the following URLs in dev:
- https://dev-doc.rust-lang.org/
- https://dev-doc.rust-lang.org/rand
- https://dev-doc.rust-lang.org/1.78.0/
- https://dev-doc.rust-lang.org/trpl
- https://dev-doc.rust-lang.org/stable/trpl
- https://dev-doc.rust-lang.org/nightly/trpl
- https://dev-doc.rust-lang.org/adv-book
- https://dev-doc.rust-lang.org/doc doesn't seem to work, I got a blank page. but it's the same behavior as prod
- https://dev-doc.rust-lang.org/stable
- https://dev-doc.rust-lang.org/nightly/

So overall looks good. I'm merging and deploying to prod as well

PR #546, opened 2024-09-02
fix(ci): don't run checks twice on PRs

MarcoIeni:
Checks run twice on PRs. This PR should fix this

image


MarcoIeni:
I'm merging this. Scream if you dislike it 👍

PR #563, opened 2024-09-04
fix(ansible): prevent apply script from crashing on mac

MarcoIeni:
Me and JD experience https://stackoverflow.com/questions/73638290/python-on-mac-is-it-safe-to-set-objc-disable-initialize-fork-safety-yes-globall

This should fix it.


jdno (approved):
🥳

PR #565, opened 2024-09-05
chore(dev-desktops): remove postfix

MarcoIeni:
I removed postfix because when updating to ubuntu 24 there was an alert related to it.
Instead of fixing it I decided to remove this package as we don't need it imo.

I want to merge this and apply to all ubuntu 22 images so that when we upgrade there's no error.


jdno (approved):
Seems fine :+1:

PR #578, opened 2024-09-25
Remove IP filter from bastion's security group

jdno:
In the past, access to the bastion host was restricted to static IP addresses belonging to the users with access. This worked well for users with a (semi-)static IP address, but introduced a lot of friction for users on most residential internet connections in Europe (e.g. when accessing the server from their home). Due to this friction, we have decided[^1] to drop the IP filter.

[^1]: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/drop.20the.20bastion.20IP.20requirement.2C.20or.20idea.20for.20a.20better.20way.2E.2E.2E

### DIFF
--- a/terraform/bastion/firewall.tf
+++ b/terraform/bastion/firewall.tf
@@ -1,41 +1,10 @@
 // The resources in this file control who has access to the bastion server.
 
-locals {
-  // Users allowed to connect to the bastion through SSH. Each user needs to
-  // have the CIDR of the static IP they want to connect from stored in AWS SSM
-  // Parameter Store (us-west-1), in a string key named:
-  //
-  //     /prod/bastion/allowed-ips/${user}
-  //
-  allowed_users = [
-    "aidanhs",
-    "joshua",
-    "pietro",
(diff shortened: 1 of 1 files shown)


jdno:
I'm going to clean up the `/prod/bastion/allowed-ips` once this has been merged and deployed.

terraform/bastion/firewall.tf
```diff
@@ -1,40 +1,10 @@
...
-    "technetos",
-    "nemo157",
-    "syphar",
-    "rylev",
-    "rylev-ip-2",
-    "jdn",
-    "guillaumegomez",
-    "marcoieni",
-  ]
-}
```
MarcoIeni, line 25: Is this the list of allowed users now?
https://github.com/rust-lang/simpleinfra/blob/ca5e95c263f4fe4392fb4de4160f21bf0551f42c/ansible/playbooks/bastion.yml#L13

MarcoIeni (approved):

PR #579, opened 2024-09-25
chore: grant Jake playground ec2 ssh permissions

MarcoIeni:


Mark-Simulacrum (approved):

terraform/team-members-access/infra-deploy-playground.tf
```diff
@@ -30,6 +30,26 @@ resource "aws_iam_group_policy" "infra_deploy_playground" {
           "arn:aws:ssm:us-west-1:890664054962:parameter/staging/ansible/all/*",
         ]
       },
+      {
+        "Effect" : "Allow",
+        "Action" : "ec2-instance-connect:SendSSHPublicKey",
+        "Resource" : [
+          "arn:aws:ec2:us-west-1:890664054962:instance/i-04db6ae41566bd1d9",
```
shepmaster, line 37: Would it make sense to do this via "tags" and give playground instances a tag? That way it works for the next playground instances for free.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-configure-IAM-role.html#eic-permissions-allow-users-to-connect-to-instances-with-specific-tags

terraform/team-members-access/infra-deploy-playground.tf
```diff
@@ -30,6 +30,26 @@ resource "aws_iam_group_policy" "infra_deploy_playground" {
...
+        ],
+        "Condition" : {
+          "StringEquals" : {
+            "ec2:osuser" : "shepmaster"
```
shepmaster, line 41: I'm not sure about this value; the docs suggest it being the user that the AMI creates (e.g. `ubuntu` for us, I think).

If it is *my* OS username, that'd be `shep`.

MarcoIeni:
applied

PR #581, opened 2024-09-26
chore: remove useless condition

MarcoIeni:


shepmaster:
For future context...

My guess is that when starting from a totally fresh installation there _isn't_ an OS user called `shep`. Therefore the `aws` command to add the SSH key has permissions to run and "succeeds", but does nothing useful.

The correct command is 

```
aws ec2-instance-connect send-ssh-public-key [...] --instance-os-user 'ubuntu' [...]
```

Followed by a 

```
./apply prod [...] -u ubuntu
```

PR #592, opened 2024-09-30
chore(bastion): allow ubuntu user

MarcoIeni:
I need this to apply ansible for the first time

### DIFF
--- a/terraform/bastion/firewall.tf
+++ b/terraform/bastion/firewall.tf
@@ -21,6 +21,7 @@ locals {
     "jdn",
     "guillaumegomez",
     "marcoieni",
+    "ubuntu",
   ]
 }
 


MarcoIeni:
I couldn't apply this because I can't set my ip twice (I tried to set my ip to the ubuntu user as well).

jdno:
We want to drop the firewall in #578 anyways, right? Should we deploy and merge that to unblock this one?

MarcoIeni:
Still can't `ssh ubuntu@bastion2.infra.rust-lang.org` D: so yeah, dropping the ip requirement will make things easier for sure

MarcoIeni:
EDIT: I managed by using the ip instead of the url

PR #594, opened 2024-10-01
free `bastion` domain

MarcoIeni:
In preparation for bastion ubuntu24 migration

PR #595, opened 2024-10-01
assign `bastion` domain to ubuntu 24 ec2

MarcoIeni:
- [ ] apply after https://github.com/rust-lang/simpleinfra/pull/594


MarcoIeni:
Can we assign the same domain to multiple ec2?

jdno:
Theoretically yes, but it won't do what you want it to do. 😅 But why can't we just move it from the old to the new host? Right now, it seems that between the two PRs there will be a moment in time where the domain is not directed at any host.

MarcoIeni:
I was worried that assigning the same domain to two ec2 will result in an error. I'll open a new pr

MarcoIeni:
closing in favor of https://github.com/rust-lang/simpleinfra/pull/598

PR #596, opened 2024-10-01
chore: update static-website action to ubuntu 24

MarcoIeni:
awscli is not present in apt anymore. In the playground we install it with snap, but snap isn't available in docker.

So we need to do this ugly thing to install awscli in ubuntu now.

I want to test this in https://github.com/rust-lang/thanks/pull/72

### DIFF
--- a/github-actions/static-websites/Dockerfile
+++ b/github-actions/static-websites/Dockerfile
@@ -1,6 +1,16 @@
-FROM ubuntu:bionic
+FROM ubuntu:24.04
 
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y git awscli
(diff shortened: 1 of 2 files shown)


MarcoIeni:
I also opened https://github.com/rust-lang/simpleinfra/issues/597 for the future

MarcoIeni:
I tested this in https://github.com/rust-lang/thanks/commit/467b5269b28f7775cb983975c924fe1c31f808d9

It's failing because of the newest aws cli. The error message is cryptic but imo the aws failed to authenticate.

MarcoIeni:
it worked after adding the default region!

MarcoIeni:
I'm merging this, we can always revert in case of issues

PR #600, opened 2024-10-02
chore(crater): create ubuntu 24 vm

MarcoIeni:
To update the crater vm, we can follow the same strategy of bastion and playground:
- create new vm
- apply ansible
- switch `crater2.infra.rust-lang.org` and `crater.infra.rust-lang.org` domains between the two VMs

Do you think this could work?
Or maybe there's something I don't know about how the bot and agents communicate with the server that would invalidate this plan?

### DIFF
--- a/terraform/crater/instance.tf
+++ b/terraform/crater/instance.tf
@@ -51,6 +51,13 @@ resource "aws_eip" "crater" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
The instance is stateful (there's at least an sqlite database involved). I would probably recommend an in-place upgrade.

PR #607, opened 2024-10-08
fix(crater): install awscli from snap

MarcoIeni:
Ansible in master fails on crater (which is now on ubuntu 24).
This is because awscli is not installed anymore in apt.
With this PR we install aws from snap instead of apt.

How to deploy this:
- uninstall awscli with ssh
- run ansible apply

A newer version of the aws cli will be pulled so maybe there are some breaking changes that we need to fix after. ([this](https://github.com/rust-lang/simpleinfra/blob/d5a2b5d10ab290868a49ee0c385db9ca0c564313/ansible/roles/docker/templates/update-images/update.sh#L33) should work)

### DIFF
--- a/ansible/roles/docker/tasks/install.yml
+++ b/ansible/roles/docker/tasks/install.yml
@@ -4,9 +4,20 @@
   apt:
     name:
       - docker.io
-      # Needed to pull images from ECR:
-      - awscli
     state: present
+
+# Install AWS to pull images from ECR
+- name: Install aws (Ubuntu < 24)
+  apt:
+    name: awscli
+    state: present
+  when: ansible_distribution_version is version('24', '<')
+- name: Install aws (Ubuntu >= 24)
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):

MarcoIeni:
Finally I found the queue empty this morning.

image


I ran

```
sudo apt-get remove awscli
```

but ubuntu couldn't locate the awscli package.

So I manually deleted the aws binary with `sudo rm /usr/local/bin/aws`

Now I'm running ansible apply.

MarcoIeni:
Ansible finished:

`crater.infra.rust-lang.org : ok=128  changed=5    unreachable=0    failed=0    skipped=90   rescued=0    ignored=1`

So I'm merging this

MarcoIeni:
image
Crater still works 🥳

PR #609, opened 2024-10-08
allow bastion2 to access crater

MarcoIeni:


MarcoIeni:
applied. now `ssh -J marcoieni@bastion2.infra.rust-lang.org marcoieni@crater.infra.rust-lang.org` works

PR #610, opened 2024-10-08
allow bastion2 to access the playground

MarcoIeni:


MarcoIeni:
applied. `ssh -J marcoieni@bastion2.infra.rust-lang.org marcoieni@play-2.infra.rust-lang.org` works now

Issue #360, opened 2023-10-30
Redirect minor versions to latest patch


jdno:
We got [a request on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Docs.20redirect.20minor.20versions.20to.20latest.20patch) to redirect requests for the documentation from the minor version (e.g. https://doc.rust-lang.org/1.65/std/boxed/struct.Box.html) to the latest patch for that version (e.g. https://doc.rust-lang.org/1.65.0/std/boxed/struct.Box.html).

The reason is that the first URL does not exist and will return `HTTP 404 Not Found`, while the latter works fine.

The CloudFront distribution is configured [here](https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/modules/release-distribution/cloudfront-doc.tf) and is already using a [Lambda router](https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/modules/release-distribution/lambdas/doc-router/index.js) to issue some redirects.

MarcoIeni:
how do we get the latest patch from the edge function?
Either we hardcode it (it seems a waste of time) or we do an HTTP call somewhere but it seems a waste of resource.

I think the easiest solution is just to append `.0` and live with it.
I don't think that docs are going to change dramatically from patch to patch.

What do you think?

MarcoIeni:
Also another doubt: we don't need to edit fastly code to make this change right? From a quick look it seems that the doc router is only on cloudfront 👍

jdno:
> how do we get the latest patch from the edge function?

This question is the reason why no one has started working on this yet. 😅

I don't think we have a good solution for this. We _could_ fetch it from the function by doing an HTTP call, but we should avoid that since it will very negatively impact response times. We _could_ hard-code the releases in the function and just look up the latest patch version for each minor version, but that would need to be updated with every release.

Doing some quick math on the GitHub releases for `rust-lang/rust`, it seems that out of ~130 releases only 34 were a patch. So it feels kinda reasonable to default to `.0`, although ideally we'd have a "you are viewing outdated documentation" banner. But I guess that's a feature request for the rustdoc team.

jdno:
> From a quick look it seems that the doc router is only on cloudfront 👍

That is correct. 👍  Only `static.crates.io` and `static.rust-lang.org` are served through Fastly right now.

MarcoIeni:
[...]

PR #612, opened 2024-10-09
Ensure patch version is provided if accessing versioned documentation

nain-F49FF806:
This is step one in addressing https://github.com/rust-lang/simpleinfra/issues/360

Have modified the router slightly

- If complete (including patch) version provided, forward as is
- If only upto minor version is provided, redirect to patch zero of that minor version

This should help in redirecting minor version documentation access to (hopefully) valid pages

Todo/Future: Consider forwarding to latest patch when only minor version provided


terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,17 @@ exports.handler = (event, context, callback) => {
...
+    // Forward patch versioned documentation as-is.
+    if (/^(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
         return callback(null, request);
     }
+    // Include patch version 0 if only minor version is provided
+    // Todo("Forward to latest patch version if possible")
+    const minor_versioned_pattern = /^(\d+)\.(\d+)\/(.*)/
+    if (minor_versioned_pattern.test(request.uri)) {
+        const patched_uri = request.uri.replace(minor_versioned_pattern, "$1.$2.0/$3");
+        return temp_redirect(patched_uri, callback);
```
nain-F49FF806, line 78: Note: Temp redirect is used so the patch version used can be updated later.

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,17 @@ exports.handler = (event, context, callback) => {
         return temp_redirect('https://www.rust-lang.org/learn', callback);
     }
 
-    // Forward versioned documentation as-is.
-    if (/^\/\d/.test(request.uri)) {
+    // Forward patch versioned documentation as-is.
+    if (/^(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
```
MarcoIeni, line 70: This regex looks good 👍
image

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,17 @@ exports.handler = (event, context, callback) => {
...
         return callback(null, request);
     }
+    // Include patch version 0 if only minor version is provided
+    // Todo("Forward to latest patch version if possible")
```
MarcoIeni, line 74: I'm not sure if we can do this.
Asked in the original issue.
If we can't let's delete this line. 👍

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,17 @@ exports.handler = (event, context, callback) => {
...
+    const minor_versioned_pattern = /^(\d+)\.(\d+)\/(.*)/
```
MarcoIeni, line 75: This regex looks good 👍
image

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
MarcoIeni: why do we want to update it later? 🤔

MarcoIeni:
Looks great 🥳
Let me check if this is all that's required and I'll deploy it to the dev environment

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,17 @@ exports.handler = (event, context, callback) => {
         return temp_redirect('https://www.rust-lang.org/learn', callback);
     }
 
-    // Forward versioned documentation as-is.
-    if (/^\/\d/.test(request.uri)) {
+    // Forward patch versioned documentation as-is.
+    if (/^(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
         return callback(null, request);
     }
+    // Include patch version 0 if only minor version is provided
```
MarcoIeni, line 73:
```suggestion
    // Include patch version 0 if minor version is provided
```

MarcoIeni:
Applied to dev-doc.rust-lang.org 👍
However https://dev-doc.rust-lang.org/1.65/std/boxed/struct.Box.html it's not working

MarcoIeni:
But https://dev-doc.rust-lang.org/1.65.0/std/boxed/struct.Box.html it's not working too.
let me check

MarcoIeni:
I redeployed the master branch and apparently those URLs are not working in dev 🤔

MarcoIeni:
Ok. Now https://dev-doc.rust-lang.org/1.65.0/std/boxed/struct.Box.html is working

MarcoIeni:
I applied your code and https://dev-doc.rust-lang.org/1.65/std/boxed/struct.Box.html isn't working.

Do you have any idea? Maybe the regex substitution doesn't work?

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
nain-F49FF806: This is in case we could update the code to get the latest patch live and direct to it. 
If we can then temp redirect would be good, as the latest patch version for a minor release can change. 

If we are happy with always redirecting to the patch zero release. Then regular redirect should do.

nain-F49FF806:
> I applied your code and https://dev-doc.rust-lang.org/1.65/std/boxed/struct.Box.html isn't working.
> 
> Do you have any idea? Maybe the regex substitution doesn't work?


Hi, I checked the request in browser's dev console.
image


I suspect the pattern has not matched due to the leading `/` indicating the root. 
Have made a commit fixing the regex accordingly. 

Could you try now?


p.s: 
  If the above is the correct reason why it didn't work then we may have to update the earlier regex also. 
  Have edited only one regex now  so we can see if it works without changing behavior of the patch versioned case.

MarcoIeni:
Deloyed 👍
https://dev-doc.rust-lang.org/1.65/std/boxed/struct.Box.html still doesn't work

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,16 @@ exports.handler = (event, context, callback) => {
...
     }
 
-    // Forward versioned documentation as-is.
-    if (/^\/\d/.test(request.uri)) {
+    // Forward patch versioned documentation as-is.
+    if (/^(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
         return callback(null, request);
     }
+    // Include patch version 0 if minor versioned documentation is requested
+    const minor_versioned_pattern = /^\/?(\d+)\.(\d+)\/(.*)/
```
MarcoIeni, line 74:
```suggestion
    const minor_versioned_pattern = /^\/(\d+)\.(\d+)\/(.*)/
```
The first `/` is always there I think

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -66,10 +66,16 @@ exports.handler = (event, context, callback) => {
         return temp_redirect('https://www.rust-lang.org/learn', callback);
     }
 
-    // Forward versioned documentation as-is.
-    if (/^\/\d/.test(request.uri)) {
+    // Forward patch versioned documentation as-is.
+    if (/^(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
```
MarcoIeni, line 70:
```suggestion
    if (/^\/(\d+)\.(\d+).(\d+)\/(.*)/.test(request.uri)) {
```
There's a `/` at the beginning, right? It was there before.

nain-F49FF806: Absolutely! I seem to have missed it. Have added it back now.

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
nain-F49FF806: You are correct! It is always [required](https://www.rfc-editor.org/rfc/rfc7230#section-5.3.1) to be present.
Have edited accordingly.

jdno:
This is something that I'd love to see added to our [rust-lang/infra-smoke-tests](https://github.com/rust-lang/infra-smoke-tests). It will probably require creating a new `TestSuite`, so I'll try to create a tracking issue with some instructions later.

MarcoIeni:
Looks great, I'm testing this one last time just to be sure 😂

MarcoIeni (approved):
This is working for me and the code looks good. 👍
Great job, thanks 🙌
I'll wait for a review from JD since CDN is quite critical.

MarcoIeni:
One note for the maintainers: I want to merge this with "squash" strategy. Is that fine or do we want to the contributors to force push?

jdno (approved):
Looks good! I'm adding it to our [smoke tests](https://github.com/rust-lang/infra-smoke-tests), but we can already go ahead and deploy it...

MarcoIeni:
nain, can you squash your commits into a single one? After that, push a new commit where you update [this](https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/accounts/legacy/releases-prod/deployed-ref#L1) sha with the sha of your first commit.

So in total, this PR should contain 2 commits.

MarcoIeni:
Applied to production 👍

MarcoIeni:
https://doc.rust-lang.org/1.65/std/boxed/struct.Box.html it's working 🎉

Thanks nain and sorry if it took us so long to review but CDNs are quite critical!

I hope to see you again in my GitHub notifications 😁

PR #616, opened 2024-10-18
dev-desktops: Install npm

aDotInTheVoid:
NPM is used by rustdoc for the rustdoc-gui tests.

I have no way to check if this is correct/works.

### DIFF
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -136,7 +136,9 @@
 
 - name: Install Node
   apt:
-    name: nodejs
+    name: 
+      - nodejs
+      - npm
     state: present
     update_cache: yes
 


ansible/roles/dev-desktop/tasks/dependencies.yml
```diff
@@ -136,7 +136,9 @@
 
 - name: Install Node
   apt:
-    name: nodejs
+    name: 
+      - nodejs
+      - npm
     state: present
     update_cache: yes
```
Noratrieb, line 143:
```suggestion
- name: Install Node
  apt:
    name: {{ item }}
    state: present
    update_cache: yes
  with_items:
   - node
   - npm
```
I think you need to wrap it in a loop (or just make two separate steps)

aDotInTheVoid: Just using a list works for the main dependency step, so I _think_ it's fine

https://github.com/rust-lang/simpleinfra/blob/9332e75d1ecdb6341b787a6b3c59df0bf908627d/ansible/roles/dev-desktop/tasks/dependencies.yml#L11-L18

Noratrieb: ah, right. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_module.html#parameter-name

jdno (approved):
Oh weird that no one complained yet that we only install Node and npm. 😅

jdno:
This has been deployed to all dev-desktops. Feel free to give it a try and let me know if you run into any issues... 🙂

PR #619, opened 2024-10-21
swap domain of bastion 1 and 2

MarcoIeni:


MarcoIeni:
applied, but `ssh marcoieni@bastion.infra.rust-lang.org` still points to the old ec2 for me

PR #620, opened 2024-10-22
Fix the Ansible playbook for dev-desktops

jdno:
The Ansible playbook for the dev-desktops did not apply cleanly anymore after a few recent changes.

- The latest minor version of the Datadog agent (`7.57.x`) fails to install with an `unexpected end of file or stream` error from dpkg.
- After installing `earlyoom` in #480, the Ansible playbook cannot be applied to staging anymore since `earlyoom` will kill the Rustup installer.
- Debugging the installation step for Rust was quite painful. The inline script with multiple steps has been refactored into separate Ansible tasks, which makes this much better.
- The previous Rustup version was from June 2021 and has been upgraded to the latest release.

### DIFF
--- a/ansible/playbooks/dev-desktop.yml
+++ b/ansible/playbooks/dev-desktop.yml
@@ -17,6 +17,7 @@
         datadog_api_key: "{{ vars_datadog_api_key }}"
         datadog_site: "datadoghq.com"
 
+        datadog_agent_version: "7.56.2"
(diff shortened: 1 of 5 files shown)


MarcoIeni:
What if we disable earlyoom for processes ran by admins?

MarcoIeni (approved):

PR #626, opened 2024-10-28
docs(bastion): explain why we have the `2` in the name of resources

MarcoIeni:
Unfortunately I can't go back to the previous name, otherwise it will destroy and recreate the resources


jdno:
You can rename the resources in the state using the [`terraform state mv`](https://developer.hashicorp.com/terraform/cli/commands/state/mv) command.

PR #628, opened 2024-10-30
Restore LLVM fix

jdno:
The prior fix for LLVM that was removed in #562 has been restored. The Python versions in the fix have been updated to match the versions that ship with Ubuntu 24.04.


ansible/roles/dev-desktop/tasks/fix_llvm_55575.yml
```diff
@@ -0,0 +1,30 @@
+---
+
+# lldb 14 fails to load Python modules due to a wrong path. The workaround for
+# this issue is to symlink the modules to the path that lldb expects.
+# https://github.com/llvm/llvm-project/issues/55575
```
MarcoIeni, line 5:
```suggestion
# https://github.com/llvm/llvm-project/issues/55575
#
# This fix can be removed once we move to llvm 16
```

PR #653, opened 2024-12-20
Set my default shell to `fish`on dev-desktops

jieyouxu:
I forgot you could do that.


MarcoIeni:
I'm applying now...

MarcoIeni:
it should be applied now 👍

jieyouxu:
Thanks @MarcoIeni, I can confirm I see 🐟

PR #654, opened 2024-12-20
Reduce the number of monitorbot tasks to 0

pietroalbini:
Monitorbot keeps failing to start up. While we decide what to do with it, avoid it from starting up.


pietroalbini:
(this is already deployed)

PR #672, opened 2025-03-11
change docs redirect

marcoieni:
fix https://github.com/rust-lang/rust/issues/77537#issuecomment-2713674385

### DIFF
--- a/terragrunt/modules/release-distribution/lambdas/doc-router/index.js
+++ b/terragrunt/modules/release-distribution/lambdas/doc-router/index.js
@@ -61,9 +61,9 @@ exports.handler = (event, context, callback) => {
     const request = event.Records[0].cf.request;
     console.log("request.uri = " + request.uri);
 
-    // Redirect `/` to the rust-lang.org website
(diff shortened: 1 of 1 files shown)


marcoieni:
I applied this in dev, so in theory dev-doc.rust-lang.org should redirect to the new url, but it doesn't 🤔

terragrunt/modules/release-distribution/lambdas/doc-router/index.js
```diff
@@ -63,7 +63,7 @@ exports.handler = (event, context, callback) => {
 
     // Redirect `/` to the rust-lang.org website
     if (request.uri === '/' || request.uri === '/index.html') {
-        return temp_redirect('https://www.rust-lang.org/learn', callback);
+        return temp_redirect('https://doc.rust-lang.org/stable/', callback);
```
marcoieni, line 66: don't hardcode the domain.

marcoieni:
this works in dev https://dev-doc.rust-lang.org/

jdno (approved):

PR #675, opened 2025-03-17
create aws account ci-prod

marcoieni:


marcoieni:
applied but I still don't see the account.

marcoieni:
after https://github.com/rust-lang/simpleinfra/pull/676 I see the account

PR #698, opened 2025-04-05
Allow the ci-mirrors repo to upload artifacts

pietroalbini:
See https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/idea.3A.20automating.20ci-mirrors

This is already deployed.

### DIFF
--- a/terraform/shared/ci-mirrors-access.tf
+++ b/terraform/shared/ci-mirrors-access.tf
@@ -26,8 +26,20 @@ resource "aws_iam_role" "foo" {
       Statement = [
         {
           Effect   = "Allow"
-          Action   = ["s3:GetObject", "s3:PutObject"]
+          Action   = "s3:GetObject"
           Resource = "${aws_s3_bucket.rust_lang_ci_mirrors.arn}/*"
+        },
+        {
+          Effect   = "Allow"
+          Action   = "s3:PutObject"
+          Resource = "${aws_s3_bucket.rust_lang_ci_mirrors.arn}/*"
+          Condition = {
+            StringEquals = {
+              // Enforce that the `if-none-match: *` header is provided when uploading a new file,
+              // ensuring CI can never override an existing file.
+              "s3:if-none-match" = "*"
+            }
+          }
         },
       ]
     })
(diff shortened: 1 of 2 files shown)


terraform/shared/ci-mirrors-access.tf
```diff
@@ -0,0 +1,35 @@
...
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Principal = {
+          Federated = aws_iam_openid_connect_provider.github_actions.arn
+        }
+        Condition = {
+          StringLike = {
+            "token.actions.githubusercontent.com:sub" = "repo:rust-lang/ci-mirrors:ref:refs/heads/gh-readonly-queue/main/*"
```
Mark-Simulacrum, line 15: nit: we might want a different branch name since this does have PutObject permissions?

terraform/shared/ci-mirrors-access.tf
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  inline_policy {
+    name = "permissions"
+    policy = jsonencode({
+      Version = "2012-10-17"
+      Statement = [
+        {
+          Effect   = "Allow"
+          Action   = ["s3:GetObject", "s3:PutObject"]
```
Mark-Simulacrum, line 29: Should we enforce never overwriting an object?

https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes-enforce.html#conditional-writes-enforce-ex1

(This is pretty recently launched but seems like a good fit for our purposes here)

pietroalbini: That's amazing! Yes, we should definitely do that, I'll update the PR probably tomorrow.

terraform/shared/ci-mirrors-access.tf
pietroalbini: i don't think we can change the branch name, it's what GitHub Merge Queues put code into. I have a different idea for how to organize CI (have the apply be done by a separate job that doesn't build any code, to reduce permissions even more), and there I can just switch to our usual environment-based authorization. I'll do so tomorrow.

pietroalbini: Switched to an environment, as per https://github.com/rust-lang/ci-mirrors/pull/2

terraform/shared/ci-mirrors-access.tf
pietroalbini: Added a policy to never overwrite the file.

Mark-Simulacrum (approved):

PR #703, opened 2025-05-15
add CORS headers to index.crates.io

jlizen:
Context: [zulip chat thread](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/CORS.20headers.20on.20https.3A.2F.2Findex.2Ecrates.2Eio/near/518212938)

Currently despite a [permissive data access policy](https://crates.io/data-access#crate-index), `index.crates.io` doesn't return CORS headers. This prevents client-side javascript lookups of versions and other metadata per crate. Such lookups are useful for other websites that are visualizing crate information.

Per our discussion in Zulip, I'm adding simple Cloudfront-injected viewer response headers to add `access-control-allow-origin = "*"`. I'm not adding `OPTIONS` headers, support for non-simple fetch headers, or other more complex CORS support. It shouldn't be relevant given that we want to support clients that are simply retrieving crate metadata via `fetch` `GET` calls.

## Testing

I didn't directly test this code, though I did copy most of it from similar files. The CORS config came directly from the [hashicorp reference](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_response_headers_policy#cors_config-1)

The instructions on testing in the README involve needing permissions to make CLI calls that I don't have. I could probably install `terragrunt` and figure out how to get it to run `terragrunt validate` if you prefer... I don't think Github CI is doing that, just formatting. Let me know!

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/fastly.toml
+++ b/terragrunt/modules/crates-io/compute-static/fastly.toml
@@ -9,4 +9,4 @@ name = "compute-static"
 service_id = ""
 
 [scripts]
-  build = "cargo build --bin compute-static --release --target wasm32-wasi --color always"
+build = "cargo build --bin compute-static --release --target wasm32-wasip1 --color always"
--- a/terragrunt/modules/crates-io/compute-static/rust-toolchain.toml
+++ b/terragrunt/modules/crates-io/compute-static/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "stable"
-targets = [ "wasm32-wasi" ]
+targets = ["wasm32-wasip1"]
--- a/terragrunt/modules/crates-io/s3-index.tf
+++ b/terragrunt/modules/crates-io/s3-index.tf
@@ -4,6 +4,14 @@ resource "aws_s3_bucket" "index" {
   versioning {
     enabled = true
   }
+
+  // Allow browsers to fetch index files from JavaScript.
+  cors_rule {
+    allowed_methods = ["GET"]
+    allowed_headers = ["*"]
+    allowed_origins = ["*"]
+    max_age_seconds = 3000
+  }
 }
 
 resource "aws_s3_bucket_policy" "index" {


terragrunt/modules/crates-io/cloudfront-index.tf
```diff
@@ -82,3 +83,32 @@ resource "aws_route53_record" "index" {
...
+    access_control_allow_methods {
+      items = ["GET"]
+    }
+
+    access_control_allow_origins {
+      items = ["*"]
+    }
+
+    # if we ever start setting CORS headers on the origin, they take precedence
+    # (currently we don't)
```
Mark-Simulacrum, line 111: There's a comment above about the cache being different based on the origin header - implying that s3 does set CORS sometimes - can you elaborate/figure out whether there's some subtle aspect we're missing here?

Mark-Simulacrum: https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html sort of implies we should be configuring s3, not necessarily CloudFront here? Though configuring CloudFront also seems fine...

jlizen: The current behavior is that no headers are set. If for some reason that were to change, this setting would pass them through. That seems wise since of somebody started setting headers on certain keys via S3 object config, they presumably had a good reason. 

Glad to update this comment to clarify if that seems like a reasonable choice!

jlizen: > sort of implies we should be configuring s3

You could do it on S3, or cloudfront, it doesn't much matter in this case. I tend to prefer to do it on the outermost layer to keep it easier to reason about. 

Cloudfront also has managed policies that handle cors, etc -
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/understanding-response-headers-policies.html#understanding-response-headers-policies-cors

Mark-Simulacrum: Yeah, let's update both comments to avoid future confusion and I can deploy then.

jlizen: Coming right up!

jlizen: I dug around a little more, just a few notes for completeness:

The comment around the s3 caching seems to have been introduced in this commit: https://github.com/rust-lang/simpleinfra/commit/1f82aef815f22e1c49315e81939a5ea202454499 (and then propagated through various migrations).

That was in 2019, which predated [cloudfront having nice CORS support](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-cloudfront-supports-cors-security-custom-http-response-headers), so yeah it was necessary to set the CORS headers on the origin. And then specifically, that commit was adding `allowed_headers = ["*"]` which means it was trying to support non-simple fetch calls.

Meanwhile our current index bucket doesn't have any CORS-related config: https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/modules/crates-io/s3-index.tf

For context, the [documentation on CORS headers for S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html#cors-eval-criteria) only talks about preflight OPTIONS requests, which we're not bothering with since we're only trying to support simple GET requests. In fact we're not even allowing through OPTIONS calls on our distribution in the first place.

Though, I'm a bit suspicious that the S3 docs are misleading... I would hope that it still returns CORS headers even on simple calls. Just because they don't need preflight checks, doesn't mean that the browser won't care about the headers it get back.

In fact, that's exactly the problem that this PR aims to solve - missing headers on simple get calls ^^

jlizen: Anyway, the TLDR is it's a bit of a nothingburger and that the comment is misleading. I'll try to clarify. I think the carried-forward config to pass through the Origin header is still probably wise, for the same reason that propagating any CORS headers that the origin sends is wise. Saves some future head-scratching if the use case changes.

jlizen: I actually thought more about it, and I think it's probably better to set the headers on the S3 origin for consistency with static.crates.io . If it were me, I'd probably put them both on the cloudfront layer, but seeing as static.crates.io is already configured one way, doesn't seem to be much point to rock that boat.

Mark-Simulacrum:
OK, pushed a slight delta in the commits (also fixing outdated fastly wasm config) but deployed this to staging index. If we get no reports of issues, will push to production index bucket sometime next week.

jlizen:
Thanks, confirming that the CORS headers are now showing up in staging and allowing simple fetch calls through.

PR #705, opened 2025-05-16
Add cvise to dev-desktops

saethlin:
cvise is a test case reducer that I want to use on fuzzer findings: https://github.com/marxin/cvise


jdno:
I've deployed this to the dev-desktops. Let me know if it works as expected. :)

saethlin:
It works! Thanks!

PR #706, opened 2025-05-17
give Urgau triagebot access

marcoieni:
https://github.com/rust-lang/team/blob/master/people/Urgau.toml is part of the triagebot team


marcoieni:
There are some uncommitted changes in the main branch, so I can't apply this for now.

PR #717, opened 2025-05-25
dev-desktop: install libxcb1-dev libxcb-render0-dev libxcb-shape0-dev libxcb-xfixes0-dev

zjp-CN:
[Jless requires them](https://github.com/PaulJuliusMartinez/jless?tab=readme-ov-file#dependencies):

> On Linux systems, X11 libraries are needed to build clipboard access if building from source. On Ubuntu you can install these using apt install.

[These libs are supported on both amd64 and arm64.](https://packages.ubuntu.com/search?keywords=libxcb&searchon=names)


marcoieni:
Applied! Let me know if it doesn't work 👍

PR #718, opened 2025-05-26
Set default shell for jhpratt

jhpratt:


marcoieni:
Applied! Let me know if it doesn't work 👍

PR #719, opened 2025-05-27
Add Terraform configuration for gha-self-hosted's images storage

pietroalbini:
As part of https://github.com/rust-lang/gha-self-hosted/issues/11 we need to host the VM images for gha-self-hosted built by CI *somewhere*. This PR sets up that somewhere in the legacy account, with one storage bucket, an IAM role for CI to upload stuff to the bucket, and a CloudFront distribution serving the content (https://gha-self-hosted-images.infra.rust-lang.org/).

I placed this in the legacy account because there is not really a suitable account, and this will be the only bit of information related to gha-self-hosted to be stored in AWS. I don't want it to live in the `ci-*` accounts, as those are meant to run untrusted CI builds.

This is already deployed.

### DIFF
--- a/terragrunt/modules/gha-self-hosted-images/cdn.tf
+++ b/terragrunt/modules/gha-self-hosted-images/cdn.tf
@@ -14,7 +14,7 @@ resource "aws_cloudfront_distribution" "cdn" {
   viewer_certificate {
     acm_certificate_arn      = module.certificate.arn
     ssl_support_method       = "sni-only"
-    minimum_protocol_version = "TLSv1.1_2016"
+    minimum_protocol_version = "TLSv1.2_2021"
   }
 
   default_cache_behavior {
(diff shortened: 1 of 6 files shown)


terragrunt/modules/gha-self-hosted-images/cdn.tf
```diff
@@ -0,0 +1,106 @@
...
+
+  enabled             = true
+  wait_for_deployment = false
+  is_ipv6_enabled     = true
+  price_class         = "PriceClass_All"
+
+  viewer_certificate {
+    acm_certificate_arn      = module.certificate.arn
+    ssl_support_method       = "sni-only"
+    minimum_protocol_version = "TLSv1.1_2016"
```
Mark-Simulacrum, line 17: Can you bump this up to something more modern? Ideally the most recently defined policy.

Mark-Simulacrum (approved):

PR #738, opened 2025-06-12
Add configuration for the rustc-perf collector machine

Kobzol:
The only missing thing is the connection to the shared RDS Postgres database. Also one thing that has to be configured manually currently are the contents of the `run.sh` benchmarking script.


marcoieni (approved):

Kobzol:
(This is already applied, created a label to make that explicit).

marcoieni:
ok, I merged it than 👍

PR #741, opened 2025-06-20
allow rustc-perf collection server to access bastion

marcoieni:


terraform/rustc-perf/dns.tf
```diff
@@ -17,3 +17,10 @@ resource "aws_route53_record" "one" {
   records = ["144.76.186.39"]
   ttl     = 300
 }
+
+output "rustc_perf_ips" {
+  value = [
+    tolist(aws_route53_record.legacy.records)[0],
+    tolist(aws_route53_record.one.records)[0]
+  ]
+}
```
marcoieni, line 26: move this to a outputs.tf file

PR #751, opened 2025-07-19
Prevent infinite redirect for doc-router `/doc[^/]`

Noratrieb:
Previously, it would just put the slice in the Location with no regards
for what was in there. If there was no slash after the `/doc`, it would
issue a relative redirect, which was still under `/doc`, causing an
infinite redirect.

- /docs/std -> s/std
- /docs/s/std -> s/s/std
- /docs/s/s/s/std -> ...

We prevent this behavior here by only applying the `/doc` redirect if
there is a slash directly after it, causing the default /stable rewrite
fallback to be applied to everything else, which will in practice lead
to nice 404 page.

This should not break anything as all such queries currently run into
these loops, including plain `/doc`.

Making a change to this core part of infrastructure that I can't reasonably test locally scared me, so I just wrote a bunch of unit tests for the thing.

### DIFF
--- a/terragrunt/modules/release-distribution/lambdas/doc-router/test.mjs
+++ b/terragrunt/modules/release-distribution/lambdas/doc-router/test.mjs
@@ -1,13 +1,18 @@
+// A test that asserts that the redirects work correctly.
+// It can be run with
+// node --test test.mjs
+
 import { test } from "node:test";
 import assert from "node:assert";
 import lambda from "./index.js";
 
 function invokeHandler(uri) {
-  return new Promise((res) => {
-    const cb = (_, response) => {
-      res(response);
+  return new Promise((resolve) => {
+    const callback = (_, response) => {
+      resolve(response);
     };
 
+    // Invoke the lambda handler with an event containing the request URI
     lambda.handler(
       {
         Records: [
@@ -21,7 +26,7 @@ function invokeHandler(uri) {
         ],
       },
       undefined,
-      cb
+      callback
     );
   });
 }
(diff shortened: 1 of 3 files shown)


Noratrieb:
I discovered this because someone else found this and posted about it, seemingly because they made a typo and forgot the URLs or something like that.

marcoieni:
thanks for this. The change itself looks good. Regarding the tests, can you document how to run them?

Noratrieb:
There is a comment at the top of the tests that mentions it, just `node --test test.mjs`

marcoieni:
I converted the PR to draft so that it's clear that it's not ready to be reviewed yet 👍

Noratrieb:
Ah no, I did make a regression test, just not as part of the table tests. I squashed the comment fix I made earlier into the commit and it should be ready now.

terragrunt/modules/release-distribution/lambdas/doc-router/test.mjs
```diff
@@ -0,0 +1,79 @@
+// A test that asserts that the redirects work correctly.
+// It can be run with
+// node --test test.mjs
+
+import { test } from "node:test";
+import assert from "node:assert";
+import lambda from "./index.js";
+
+function invokeHandler(uri) {
+  return new Promise((res) => {
```
marcoieni, line 10:
```suggestion
  return new Promise((resolve) => {
```

terragrunt/modules/release-distribution/lambdas/doc-router/test.mjs
```diff
@@ -0,0 +1,79 @@
...
+    const cb = (_, response) => {
+      res(response);
+    };
+
```
marcoieni, line 14:
```suggestion

    // Invoke the lambda handler with an event containing the request URI
```

terragrunt/modules/release-distribution/lambdas/doc-router/test.mjs
```diff
@@ -0,0 +1,79 @@
...
+// node --test test.mjs
+
+import { test } from "node:test";
+import assert from "node:assert";
+import lambda from "./index.js";
+
+function invokeHandler(uri) {
+  return new Promise((res) => {
+    const cb = (_, response) => {
+      res(response);
```
marcoieni, line 12:
```suggestion
      resolve(response);
```

marcoieni:
also it would be great if you could run the test in CI.

Noratrieb:
_oh_, this was just changed on master!

marcoieni (approved):

marcoieni:
applied, thanks!

PR #754, opened 2025-07-23
Give tiif fish shell

tiif:


tiif:
In my dev desktop, ``echo $SHELL`` is still showing ``/usr/bin/bash`` :thinking:

tiif:
It works now, thank you! :heart:

PR #768, opened 2025-09-03
fix(deploy): pipe deploy key directly to ssh-add to avoid temp file h…

AryzXploit:
Pipe deploy key directly into ssh-add to avoid writing it on disk (@pietroalbini).

### DIFF
--- a/setup-deploy-keys/src/deploy.rs
+++ b/setup-deploy-keys/src/deploy.rs
@@ -1,6 +1,5 @@
 use std::env;
-use std::fs;
-use std::io::{Read, Write};
+use std::io::Write;
 use std::os::unix::net::UnixStream;
 use std::os::unix::prelude::*;
 use std::process::{Command, Stdio};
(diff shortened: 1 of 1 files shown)


setup-deploy-keys/src/deploy.rs
```diff
@@ -1,51 +1,32 @@
...
-        .unwrap();
-
-    decode
-        .stdin
-        .take()
-        .unwrap()
-        .write_all(key.as_bytes())
-        .unwrap();
+    // decode langsung via crate base64
+    let decoded = base64::decode(&key_b64).expect("invalid base64 key");
```
pietroalbini, line 19: We can't use the base64 crate here. This script is meant to be compiled directly with `rustc`, with no crates.io dependencies.

setup-deploy-keys/src/deploy.rs
```diff
@@ -1,57 +1,42 @@
...
 
-    let path = "_the_key";
-    fs::write(path, key).unwrap();
-    fs::set_permissions(path, fs::Permissions::from_mode(0o600)).unwrap();
-    run(Command::new("ssh-add")
-        .arg(path)
-        .env("SSH_AUTH_SOCK", socket));
-    fs::remove_file(path).unwrap();
+    ssh_add.stdin.as_mut().unwrap().write_all(&decoded).unwrap();
+    ssh_add.wait().unwrap();
```
pietroalbini, line 30: We should check the status of the command to ensure it was successful.

setup-deploy-keys/src/deploy.rs
```diff
@@ -1,57 +1,42 @@
...
     while UnixStream::connect(socket).is_err() {
         std::thread::sleep(std::time::Duration::from_millis(5));
     }
 
-    let mut decode = Command::new("base64")
-        .arg("-d")
+    // decode langsung via crate base64
+    let decoded = base64::decode(&key_b64).expect("invalid base64 key");
+
+    // pipe ke ssh-add via stdin
```
pietroalbini, line 21: I'm having a bit of trouble understanding this comment 😅

pietroalbini (commented):
Left some comments. CI is failing too.

setup-deploy-keys/src/deploy.rs
AryzXploit: I've removed the base64 crate usage and replaced it with a system base64 -d pipe into ssh-add. Also cleaned up the comment for clarity.

AryzXploit:
Any updates?

AryzXploit:
helo

zulfff:
Hello

PR #799, opened 2025-10-31
Add foundation domains access for foundation folks

emilyalbini:
This PR migrates the Route53 access given to foundation staff in https://github.com/rust-lang/simpleinfra/blob/master/terraform/team-members-access/foundation.tf to IAM Identity Center, and adds Walter to that group.

This does *not* remove the old IAM User based system, we'll clean it up later.


pietroalbini:
This is already deployed.

PR #802, opened 2025-11-11
release-distribution: lower cloudfront weight

marcoieni:
The Cloudfront bill is growing too much.

The downside of this is that the 0.4% of our users will get more cache misses than before.

I discussed this with JD.
I will apply this tomorrow morning.


terragrunt/modules/release-distribution/cloudfront-static.tf
```diff
@@ -91,6 +91,11 @@ resource "aws_cloudfront_distribution" "static" {
     bucket          = data.aws_s3_bucket.logs.bucket_regional_domain_name
     prefix          = "cloudfront/${var.static_domain_name}"
   }
+
+  tags = {
+    project     = "release-distribution"
+    environment = var.environment
+  }
```
marcoieni: I added this to see a cost breakdown in the aws console

PR #807, opened 2025-11-13
docs-rs: implement rate limiting in cloudfront

marcoieni:
Generated with AI starting from the docs.rs nginx config. Didn't review yet. But `terraform plan` works

### DIFF
--- a/terraform/docs-rs/cloudfront.tf
+++ b/terraform/docs-rs/cloudfront.tf
@@ -75,6 +75,56 @@ resource "random_password" "origin_auth" {
   special = false
 }
 
+// WAF Web ACL for rate limiting
+// Porting nginx rate limiting: 10 req/s with burst of 50
+// WAF uses 5-minute windows, so 10 req/s = 3000 req/5min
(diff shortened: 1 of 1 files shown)


syphar (commented):
after thinking about it: 

It seems like the AWS WAF is running _before_ the cache, so the rate limit rules would apply before too. 

Right now, we only check rate limits in the backend, so only uncached requests. 

Seeing the difference from the stats (average of 100k rpm on cloudfront, 30k on the backend), at the very least we have to adapt the rate limit accordingly. 

But: 
The only reason for the rate limit (for me) is protecting the backend. Cached requests can go as high as they want. 

So I wonder if we should choose another approach generally, leaving cached requests open, and only rate limiting the origin requests. 

Would we have an ALB anyways? Then we could use a regional WAF in front of it? 



Even for malicious cases I mostly see a problem when these requests hit the origin, not before.

marcoieni:
> Would we have an ALB anyways?

if we need one, yes!

PR #811, opened 2025-11-19
add fastly CDN for docs-rs via `fastly.docs.rs` domain

marcoieni:
This adds the `fastly-docs.rs` domain, which goes through the fastly CDN.
we aren't changing the behavior of `docs.rs` with this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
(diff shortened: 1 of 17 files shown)


marcoieni:
When applying I got the error 

```
Error: 400 - Bad Request:
│
│     Title:  The domain you entered was not found on one of your services
│     Detail: Please add fastly-docs.rs to an active service to begin TLS enablement.
```

PR #827, opened 2025-11-25
docs-rs: switch cdn traffic to fastly

marcoieni:


terraform/docs-rs/fastly.tf
```diff
@@ -86,3 +86,23 @@ resource "aws_route53_record" "fastly_domain" {
...
+
+resource "aws_route53_record" "webapp_apex_aaaa" {
+  zone_id = data.aws_route53_zone.webapp.id
+  name    = local.domain_name
+  type    = "AAAA"
+  ttl     = 60
+
+  # You can get these IPs from: fastly.com -> security -> TLS Management -> IPs -> Domains -> fastly.docs.rs
+  records = ["2a04:4e42::347", "2a04:4e42:200::347", "2a04:4e42:400::347", "2a04:4e42:600::347"]
+}
```
kailan: Fastly doesn't really change IPs but we'd still recommend fetching them dynamically.

```suggestion

data "fastly_tls_configuration" "docs_rs_tls" {
  id = module.fastly_tls_subscription_globalsign_docs_rs.tls_configuration_id
}

resource "aws_route53_record" "webapp_apex_a" {
  zone_id = data.aws_route53_zone.webapp.id
  name    = local.domain_name
  type    = "A"
  ttl     = 60

  records = [for dns_record in data.fastly_tls_configuration.docs_rs_tls.dns_records : dns_record.record_value if dns_record.record_type == "A"]
}

resource "aws_route53_record" "webapp_apex_aaaa" {
  zone_id = data.aws_route53_zone.webapp.id
  name    = local.domain_name
  type    = "AAAA"
  ttl     = 60

  records = [for dns_record in data.fastly_tls_configuration.docs_rs_tls.dns_records : dns_record.record_value if dns_record.record_type == "AAAA"]
}
```

You'd also need to add an output to `fastly-tls-subscription` like so:

```
output "tls_configuration_id" {
  value = fastly_tls_subscription.subscription.configuration_id
}
```

Untested but this makes sense to me based on the Fastly provider docs

Issue #820, opened 2025-11-24
Migrate AWS ECS apps to ECS Express


marcoieni:
Amazon recently launched [ECS Express](https://aws.amazon.com/blogs/aws/build-production-ready-applications-without-infrastructure-complexity-using-amazon-ecs-express-mode/), which promises to reduce the amount of terraform code you need to write to provision a service (at no additional cost).

We should try to migrate our apps, starting from the simplest ones to that and see if it would be beneficial to us.

PR #830, opened 2025-11-27
feat: Migrate `fastly-exporter` to AWS ECS Express

amustaque97:
Closes: https://github.com/rust-lang/simpleinfra/issues/820


marcoieni:
from a first look it doesn't look that much simpler 😅
What do you think?

marcoieni:
So you think aws ecs express is not easily extensible?

Since the terraform code increased I'm wondering if it's suitable even for simple services.

Btw most of our services are quite simple. Just docker container and postgres

amustaque97:
> > Having said that, the limitations of ECS Express might still mean that it'd be better for us to improve/update the existing ECS modules...
> 
> Yeah, I agree, maybe it's better to create our own modules. Thank you for the investigation by the way, Mustaque!

Pleasure is all mine! Let me know what the next steps could be and whether community contributions are possible. I would like to participate 🙌🏻

bryantbiggs:
FWIW - there are already a number of well maintained modules for the custom, bespoke modules implemented here https://github.com/rust-lang/simpleinfra/tree/master/terraform/shared/modules

the major downside to rolling your own modules is that you will encounter and have to resolve issues on your own instead of relying on a community. For example, ECS Express Service has a number of issues and the implemenation as its written here is not entirely correct. See warning note we've added to the `terrafrom-aws-module` 
https://github.com/terraform-aws-modules/terraform-aws-ecs/pull/389#issuecomment-3745995128

> from a first look it doesn't look that much simpler 😅

Nor [cheaper](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/express-service-overview.html#express-service-pricing). If you have multiple services that utilize a load balancer, ECS Express Service is not going to be a cheaper solution. It will be cheaper and easier to configure a single, shared load balancer and n-number of ECS services

terraform/fastly-exporter/main.tf
```diff
@@ -6,104 +10,144 @@ data "aws_ssm_parameter" "fastly_api_token" {
...
 
-module "ecs_task" {
-  source = "../shared/modules/ecs-task"
-
-  name   = local.name
-  cpu    = 256
-  memory = 512
+resource "aws_iam_role_policy_attachment" "infrastructure" {
+  role       = aws_iam_role.infrastructure.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSInfrastructureRolePolicyForVolumes"
```
bryantbiggs: this is incorrect 


```suggestion
  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSInfrastructureRoleforExpressGatewayServices"
```

terraform/fastly-exporter/main.tf
```diff
@@ -6,104 +10,144 @@ data "aws_ssm_parameter" "fastly_api_token" {
...
 
-  log_retention_days = 7
-  ecr_repositories_arns = [
-    # This repository does not exist, since we're pulling the imge directly from GitHub.
-    # But the task module cannot be applied without providing at least one ARN here.
-    "arn:aws:ecr:us-west-1:890664054962:repository/fastly-exporter"
+# Wait for IAM roles to propagate
+resource "time_sleep" "wait_for_iam" {
+  depends_on = [
+    aws_iam_role_policy.execution,
```
bryantbiggs: this is redundant - if you are waiting on the attachment then you are already implicitly waiting on the role. 

more importantly, you shouldn't need this `time_sleep` resource or the explicit `depends_on`

PR #833, opened 2025-12-02
delete docs-rs cloudfront cdn

marcoieni:
We migrated to faslty.

TODO:
- [x] is docs-rs relying on the env var `DOCSRS_STATIC_CLOUDFRONT_ID`? Probably this doesn't matter because the ECS is not being in used in production, right?
- [x] static is still in use
- [x] docs-rs is still invalidating the other distribution


marcoieni:
applied!

PR #844, opened 2025-12-07
docs-rs: drop fastly-testing subdomain

syphar:
It's bad for search engines to have two domains serve the same content.


marcoieni:
all good now.
I deleted the subscription manually from https://manage.fastly.com/network/subscriptions
and removed the subscription from the terraform state.

PR #851, opened 2025-12-11
increase crates-io index fastly cdn weight

marcoieni:
continue to increase fastly weight. Started in https://github.com/rust-lang/simpleinfra/pull/850


Turbo87 (approved):
:shipit:

marcoieni:
applied!

PR #852, opened 2025-12-12
move crates-io index traffic to fastly

marcoieni:
nobody is complaining about how the crates-io fastly index cdn works, so we can move traffic there to save aws credits.

r? @Turbo87


Turbo87 (approved):
should we keep a small percentage on cloudfront to ensure that we'll notice when it stops working for some reason?

I'm fine with either decision though.

PR #853, opened 2025-12-12
increase hardcoded TTL for static.crates.io

syphar:
Increase hardcoded CDN TTL for static.crates.io. 

We're actively sending purges when things change, so they don't have to expire at all. 

See also: [#t-crates-io > static.crates.io / CDN caching?](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/static.2Ecrates.2Eio.20.2F.20CDN.20caching.3F/with/563294377)


marcoieni (approved):
discussed in [#t-crates-io > static.crates.io / CDN caching? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/static.2Ecrates.2Eio.20.2F.20CDN.20caching.3F/near/563287237)

PR #903, opened 2026-01-16
create fastly weighted dns records

marcoieni:


marcoieni:
when applied on staging, I got this error when visiting staging.crates.io, so we should be careful for prod

image

PR #981, opened 2026-02-12
bastion: switch AMI to Ubuntu 24.04

marcoieni:
I want to recreate the bastion instance when changing this AMI.
To do so, I removed temporarely:

```
  lifecycle {
    # Don't recreate the instance automatically when the AMI changes.
    ignore_changes = [ami]
  }
```
and applied.

### DIFF
--- a/terragrunt/modules/bastion/instance.tf
+++ b/terragrunt/modules/bastion/instance.tf
@@ -41,7 +41,12 @@ data "aws_ami" "ubuntu" {
 
   filter {
     name   = "name"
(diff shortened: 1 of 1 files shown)


terragrunt/modules/bastion/instance.tf
```diff
@@ -54,7 +59,6 @@ resource "aws_instance" "bastion" {
   ami                     = data.aws_ami.ubuntu.id
   instance_type           = "t3a.nano"
   ebs_optimized           = true
-  disable_api_termination = true
```
marcoieni: removing this option so that I can recreate the instance from terraform

PR #1019, opened 2026-03-10
provision fastly cdn for new docs-rs environment

marcoieni:
- [ ] port https://github.com/rust-lang/simpleinfra/pull/1045


syphar (commented):
why would we duplicate the code for the compute module? Or would the old one be removed later?

syphar (commented):
btw, I also see us enabling the firewall for docs.rs, but later.

PR #1039, opened 2026-04-26
Add a few more packages

jyn514:
- bat
- delta
- some random perl dependency. it's real small and doesn't break anything, i promise.

### DIFF
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -33,6 +33,9 @@
       - python3-pip
       - quota
       - ripgrep
+      - batcat
+      - git-delta
+      - libfile-which-perl #  Used by jyn's editor scripts
       - rustfilt
       - "linux-tools-{{ kernel.stdout }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/dev-desktop/tasks/dependencies.yml
```diff
@@ -33,6 +33,9 @@
       - python3-pip
       - quota
       - ripgrep
+      - batcat
```
marcoieni, line 36:
```suggestion
      - bat
```
While applying, I got this error:
```
fatal: [dev-desktop-staging.infra.rust-lang.org]: FAILED! => {"changed": false, "msg": "No package matching 'batcat' is available"}
```

On ubuntu 24 I think they renamed the package

https://github.com/sharkdp/bat#on-ubuntu-using-apt

This works:
```
docker run -it ubuntu:24.04 bash
apt update
apt install bat
```

ansible/roles/dev-desktop/tasks/dependencies.yml
marcoieni, line 23: delta was already present!

PR #1040, opened 2026-04-27
fastly: return 429 when rate limiting with ngwaf

ubiratansoares:
It seems that we should configure the status code returned by this waf rule. Since we are rate-limiting, clients will expect HTTP 429 instead of HTTP 406, which is the default for ngwaf workspace rules.

screenshot-2026-04-27 at 18 28 29

Also worths checking :

<details>
<summary>terragrunt plan</summary>

```
Terraform will perform the following actions:
  # fastly_ngwaf_workspace_rule.webapp_per_ip_rate_limit will be updated in-place
  ~ resource "fastly_ngwaf_workspace_rule" "webapp_per_ip_rate_limit" {
        id              = "69981dbcaab0692d53dedd17"
        # (6 unchanged attributes hidden)
      ~ action {
          ~ response_code     = 0 -> 429
            # (5 unchanged attributes hidden)
        }
        # (2 unchanged blocks hidden)
    }
```

</details>

### DIFF
--- a/terragrunt/modules/crates-io/fastly-webapp-ngwaf.tf
+++ b/terragrunt/modules/crates-io/fastly-webapp-ngwaf.tf
@@ -97,8 +97,9 @@ resource "fastly_ngwaf_workspace_rule" "webapp_per_ip_rate_limit" {
 
   # If the workspace mode isn't "block", then this rule is not enforced.
   action {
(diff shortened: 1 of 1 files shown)


ubiratansoares:
Applied!

PR #1042, opened 2026-04-29
disable kernel module

marcoieni:


emilyalbini:
This was done as a quick mitigation for CVE-2026-31431

PR #1059, opened 2026-05-19
fix(docs-rs): detect when client goes directly to shield

marcoieni:
## AI disclosure

I used GPT5.5 medium thinking with the codex harness to generate this change. I reviewed its output.

### DIFF
--- a/terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
+++ b/terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
@@ -79,22 +79,28 @@ fn main(mut req: Request) -> Result<Response, Error> {
         }
     }
 
-    if shield.target_is_origin() {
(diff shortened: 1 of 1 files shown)


terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
```diff
@@ -79,22 +79,28 @@ fn main(mut req: Request) -> Result<Response, Error> {
...
+        .expect("failed to get origin auth from secret store")
+        .plaintext();
+    let origin_auth = origin_auth.as_ref();
+
+    let request_is_from_edge_pop = !shield.response_is_for_client()
+        && req
+            .get_header(&X_ORIGIN_AUTH)
+            .is_some_and(|header| header.as_bytes() == origin_auth);
+
+    req.set_header(X_ORIGIN_AUTH, origin_auth);
```
marcoieni, line 94: are we leaking something here?

marcoieni:
Denis will raise a fix himself

PR #1060, opened 2026-05-19
fastly/docs.rs: use custom header between edge & shield to fix detection

syphar:


terraform/docs-rs/fastly-compute-docs-rs/src/shield.rs
```diff
@@ -72,6 +80,13 @@ impl Context {
                     err
                 )
             })?);
+
+            // we're setting a custom header on the request, so we can differentiate requests
+            // coming from fastly edge nodes from request coming from users.
+            //
+            // This could be improved at some point by using a signature here, and checking it
+            // later.
+            req.set_header(X_DOCSRS_REQUEST_SOURCE, "EDGE");
```
marcoieni: can't a malicious user set this header?
Maybe can remove this header when receiving the request from the user?

marcoieni:
do we strip this header before sending it to the origin?

terraform/docs-rs/fastly-compute-docs-rs/src/shield.rs
syphar: In my mind: 
what would happen in this case? 

from what I see, the malicious user: 
- might get rate limited earlier 
- might have a broken docs.rs (when we start using `f-forwarded-host`

Or am I missing something?

syphar:
> do we strip this header before sending it to the origin?

not right now, but why does that matter?

terraform/docs-rs/fastly-compute-docs-rs/src/shield.rs
syphar: thought about it again, you're right, this should be signed or a secret. 

( and  yes, the origin auth value could be reused)

syphar: can you recheck?

marcoieni:
> > do we strip this header before sending it to the origin?
> 
> not right now, but why does that matter?

Sorry, I should have clarified. If we strip this header when we receive the request, we make sure that WE are the one setting this header, not the user.

syphar:
> > > do we strip this header before sending it to the origin?
> 
> > 
> 
> > not right now, but why does that matter?
> 
> 
> 
> Sorry, I should have clarified. If we strip this header when we receive the request, we make sure that WE are the one setting this header, not the user.

So the current updated impl should be good?

marcoieni (approved):

PR #1062, opened 2026-05-29
Tag `manifests.txt` with surrogate key for purging

### LINKED ISSUE #828: Setup Fastly surrogate key for purging manifests.txt
We've noticed that the manifests.txt cache on Fastly is sometimes outdated after a new nightly is published. This is purge by promote-release here (https://github.com/rust-lang/promote-release/blob/44a5f66b1e0b9d780a8737abcbe41587e952dbb0/src/fastly.rs#L19, https://github.com/rust-lang/promote-release/blob/44a5f66b1e0b9d780a8737abcbe41587e952dbb0/src/main.rs#L611). That code generally speaking works for CloudFront because CloudFront supports path-based invalidation. Fastly requires 'tags' to be set on URLs at request time, and then those are invalidated later: https://www.fastly.com/documentation/reference/api/purging/#purge-tag

We will need to extend https://github.com/rust-lang/simpleinfra/blob/5fe9fb71e3f4023d64d620700ec1993e989f2d94/terragrunt/modules/release-distribution/fastly-static.tf#L82-L90 to add a tag (surrogate key) to the manifests.txt object and wire up purging for it. That tag should be `manifeststxt` per https://github.com/rust-lang/promote-release/blob/44a5f66b1e0b9d780a8737abcbe41587e952dbb0/src/fastly.rs#L48.

cc https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/manifests.2Etxt.20CDN.20outdated/with/560117719

Turbo87:
`promote-release` purges the `manifeststxt` surrogate key on every release, but nothing in the Fastly VCL tagged the https://static.rust-lang.org/manifests.txt response with that key, so the purge matched zero objects and the cached copy survived until TTL.

This adds a fetch snippet that tags `/manifests.txt` responses with the `manifeststxt` key so the existing purge takes effect.

/cc @Mark-Simulacrum @rust-lang/infra 

### Related

- Resolves https://github.com/rust-lang/simpleinfra/issues/828
- https://github.com/dependabot/dependabot-core/issues/13583
- https://github.com/dependabot/dependabot-core/pull/13604
- https://github.com/renovatebot/renovate/pull/43672

### DIFF
--- a/terragrunt/modules/release-distribution/fastly-static.tf
+++ b/terragrunt/modules/release-distribution/fastly-static.tf
@@ -104,6 +104,19 @@ resource "fastly_service_vcl" "static" {
     VCL
   }
 
+  # When a new release is published, promote-release purges the `manifeststxt`
+  # surrogate key so that `manifests.txt` updates promptly instead of lagging
+  # until TTL. This snippet tags the response with that key at request time.
+  snippet {
+    name    = "set cache key for manifests.txt"
+    type    = "fetch"
+    content = <<-VCL
+      if (req.url ~ "^\/manifests\.txt$") {
+        set beresp.http.Surrogate-Key = "manifeststxt";
+      }
+    VCL
+  }
+
   snippet {
     name    = "redirect rustup.sh to rustup.rs"
     type    = "error"


marcoieni (approved):

marcoieni:
applied!

PR #1066, opened 2026-06-02
fastly: allow temp ngwaf bypass for some users

ubiratansoares:
Work around for https://github.com/rust-lang/crates.io/issues/13695 and related issues


ubiratansoares:
Applied for both `staging` and `prod`

PR #1067, opened 2026-06-03
fastly: Cache crates.io webapp responses per origin headers

Turbo87:
The webapp Fastly service ran `return(pass)` in `vcl_recv` for every request, so it never cached anything. The comment said this was to avoid caching the API, but the same origin (`crates-io.herokuapp.com`) also serves the SvelteKit frontend, including immutable static assets that set `Cache-Control: public, max-age=315360000`. Those assets were never cached on the Fastly path, while the CloudFront distribution for the same origin cached them normally.

This removes the blanket pass so Fastly honors the origin's `Cache-Control`, matching CloudFront. `default_ttl = 0` stays, so responses without caching headers and per-user `no-store` responses are still not cached.

---

Created with the help of Claude Code, which:
- traced the missing cache hits to the unconditional `return(pass)` by inspecting Fastly vs CloudFront headers,
- reviewed the git history to understand why the snippet was added, and
- made the VCL change.

### DIFF
--- a/terragrunt/modules/crates-io/fastly-webapp.tf
+++ b/terragrunt/modules/crates-io/fastly-webapp.tf
@@ -86,16 +86,6 @@ resource "fastly_service_vcl" "webapp" {
     VCL
   }
 
-  # Pass all requests to origin (no caching for API)
-  snippet {
-    name    = "pass all requests"
-    type    = "recv"
-    content = <<-VCL
-      # Don't cache any requests - pass directly to origin
-      return(pass);
-    VCL
-  }
-
   # Handle HSTS headers if strict_security_headers is enabled
   dynamic "snippet" {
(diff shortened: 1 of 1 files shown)


marcoieni:
applied in staging, let me know how testing goes

Turbo87:
```
x-served-by: cache-fra-eddf8230101-FRA, cache-fra-etou8220020-FRA
x-cache: MISS, HIT
x-cache-hits: 0, 3
```

this is what I'm now seeing for our static assets, while the regular API responses are still `MISS, MISS`, as intended 👍

marcoieni:
should I deploy to prod?

Turbo87:
yes, please

marcoieni:
done!

PR #1070, opened 2026-06-03
create new cert for prev and beta

marcoieni:
The old certificate also had `www.rust-lang.org` which is now in github actions.
So I create a new certificate just for these two. This PR solves an error that AWS sent via email to us.

## AI disclosure

I used GPT5.5 with the codex harness to generate this change. I reviewed its output and changed it where necessary.

### DIFF
--- a/terraform/prev-rust-lang-org/_terraform.tf
+++ b/terraform/prev-rust-lang-org/_terraform.tf
@@ -22,3 +22,8 @@ terraform {
 provider "aws" {
   region = "us-west-1"
 }
+
+provider "aws" {
+  alias  = "us-east-1"
+  region = "us-east-1"
+}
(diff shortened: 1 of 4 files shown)


terraform/prev-rust-lang-org/_terraform.tf
```diff
@@ -22,3 +22,8 @@ terraform {
 provider "aws" {
   region = "us-west-1"
 }
+
+provider "aws" {
+  alias  = "us-east-1"
+  region = "us-east-1"
+}
```
marcoieni, line 29: all certificates need to be in this region.

marcoieni:
I also deleted the cloudfront distribution for www.rust-lang.org (ID E33OQ1UL0CLA1S - dq70m0d2ic23x.cloudfront.net)
image
image
image
image

marcoieni:
I also deleted the old certificate (0633f4b2-8a1f-46f8-a2d3-184c461a2eb8) which is unused
image

PR #1079, opened 2026-06-12
crates-io: Forward index.crates.io request logs to Datadog

Turbo87:
Adds a `logging_datadog` block to the `index.crates.io` Fastly service so its request logs reach Datadog, the same way the webapp and static services already do.

The block reuses the existing module-scoped `datadog-token` SSM parameter, so I think no new credentials should be needed. The existing S3 request logging (`/fastly-requests/index.crates.io/`) is left untouched and keeps flowing alongside the new Datadog stream.

/cc @rust-lang/infra


ubiratansoares (approved):
Applied to staging and prod!

PR #1082, opened 2026-06-15
fastly: remove IN workaround

marcoieni:
revert https://github.com/rust-lang/simpleinfra/pull/1066/ since the issue was resolved by fastly


ubiratansoares (approved):
Thanks!

PR #1084, opened 2026-06-17
assign names to fastly logs in datadog

marcoieni:


terragrunt/modules/crates-io/fastly-index.tf
```diff
@@ -29,6 +29,12 @@ resource "fastly_service_vcl" "index" {
   logging_datadog {
     name  = "datadog-index-${var.index_domain_name}"
     token = data.aws_ssm_parameter.datadog_token.value
+
+    format = templatefile("${path.module}/fastly-log-format.tftpl", {
```
Turbo87, line 33: what's the reason for the custom log format? what are the advantages?

terragrunt/accounts/legacy/datadog-fastly/terragrunt.hcl
```diff
@@ -29,5 +39,15 @@ inputs = {
...
     },
+    "zEKt1p82mWliIlbTQh5Hl1" = {
+      env     = "staging"
+      app     = "crates-io"
+      service = "staging-crates-io"
+    },
+    "k8I4J4zBUYdoQ0oJt98df7" = {
+      env     = "staging"
+      app     = "crates-io"
+      service = "index.staging.crates.io"
```
Turbo87, line 50: shouldn't this use dashes too for consistency?

terragrunt/accounts/legacy/datadog-fastly/terragrunt.hcl
```diff
@@ -9,6 +9,16 @@ include {
 
 inputs = {
   services = {
+    "yEvEoHsDVsUDRLd6ZTxipI" = {
+      env     = "prod"
+      app     = "crates-io"
+      service = "crates-io"
```
Turbo87, line 15: as I mentioned in the other place: why are we overriding the service IDs here? what are the pros/cons? is this considered best practice?

marcoieni:
this is not needed because we can filter by the host.

PR #1086, opened 2026-06-25
crates.io: Encode all `+` characters in static-router URI

Turbo87:
`static-router.js` used `request.uri.replace("+", "%2B")`, which with a string argument only replaces the first `+`. The Fastly Compute service uses Rust `str::replace()`, which replaces all occurrences. This switches CloudFront to a global regex to match.

The `cloudfront-js-1.0` runtime is ES5.1-based and doesn't support `replaceAll()`, so the global regex form is used.

r? @rust-lang/infra 

### Related
- https://github.com/rust-lang/crates.io/issues/4891


ubiratansoares (approved):

ubiratansoares:
Just applied to stg and prod

PR #1087, opened 2026-06-26
crates-io/compute-static: Rename build target to `wasm32-wasip1`

Turbo87:
The `wasm32-wasi` target was renamed to `wasm32-wasip1` upstream and removed in Rust 1.84. 349c667 migrated `fastly.toml` and `rust-toolchain.toml`, but missed `.cargo/config.toml`, which still pinned the old name and broke `cargo metadata` for the current toolchain.

r? @rust-lang/infra


ubiratansoares (approved):

ubiratansoares:
Applied to stg and prod with this new target

PR #1089, opened 2026-06-26
crates-io/compute-static: Upgrade fastly to 0.13.0

Turbo87:
Upgrades `fastly` and `log-fastly` from 0.9.1 to 0.13.0 in the `compute-static` function, one minor version at a time. Each step is its own commit, and for every step I read the crate's `CHANGELOG.md` and checked the changes against what we actually use.

Two source changes were required across the whole range:

- **0.11.0** made `Request::get_url_mut` return a smart pointer instead of a `&mut Url`, so the three call sites now bind it with `let mut url`.
- **0.12.0** changed `get_tls_cipher_openssl_name` and `get_tls_protocol` to return `Result<Option<&str>, Utf8Error>` with a non-`'static` borrow. We collapse both the UTF-8 error and the absent case to `None` and store owned `String`s in `TlsDetails`, matching how the sibling `HttpDetails` fields already work.

Everything else was assessed as not relevant:

- **0.10.0** `with_header` set to append: we set `LOCATION` once on a fresh response, so append and set are identical. No direct `http` dependency, and the removed `limits` / Object Store APIs are unused.
- **0.11.0** stricter `private` / `no-store` cache-control parsing: this only differs from the old behavior for headers where those words appear inside another token, which our `Cache-Control` values (`public, max-age=...`, or none) never do.
- **0.13.0** default `stale-if-error`: a window is only established by an explicit `stale-if-error` directive (Fastly does not derive it from `max-age`), which our S3 responses do not send, and it never substitutes for a `5xx`, so the primary/fallback logic is unchanged.

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/Cargo.lock
+++ b/terragrunt/modules/crates-io/compute-static/Cargo.lock
@@ -197,9 +197,9 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "fastly"
-version = "0.12.1"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "531e4c3df48350d9f4fc95b4deaf87fd29820336b7926bb84bf460457c2a126b"
+checksum = "62012ce6544fe774aa10ac4fb6b4715f8b0e3ffa5ff32c6a46b24cd625900493"
 dependencies = [
  "anyhow",
  "bytes",
@@ -210,7 +210,6 @@ dependencies = [
  "fastly-sys",
  "http",
  "itertools",
- "lazy_static",
  "mime",
  "serde",
  "serde_json",
@@ -225,9 +224,9 @@ dependencies = [
 
 [[package]]
 name = "fastly-macros"
-version = "0.12.1"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc2aef5f9690b04c8890f9a54ddb591b12b9779ec25ee0e572d207106e52e3d8"
+checksum = "23716aaac785201a3a7744c4eb929fd3def319899ed3b5c5cf00e9f2fa24625d"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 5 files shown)


marcoieni:
sorry there are some conflicts, can you resolve them?

Turbo87:
alright, rebased and conflicts fixed

marcoieni (approved):

marcoieni:
in staging I get the following error (visible in the fastly tailing logs):
```

2026-07-06 09:33:06
ce97fc9b
stderr
thread 'main' (1) panicked at /Users/marco/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fastly-0.13.0/src/config_store.rs:32:23:
2026-07-06 09:33:06
ce97fc9b
stderr
could not open config store `compute_static`: config store could not be found
2026-07-06 09:33:06
ce97fc9b
stderr
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2026-07-06 09:33:06
ce97fc9b
stderr
{"pop":"LIN","level":"error","diagnostic":"instance_error","error":"entered_unreachable","trap":{"trap_code":"wasm trap: wasm `unreachable` instruction executed","display_reason":"wasm trap: wasm `unreachable` instruction executed"}}
```

Turbo87:
can you bisect which commit is causing this?

marcoieni:
codex likely found the issue:

> I found the staging failure. It came from the Fastly Rust SDK upgrade from `fastly 0.9.1` to `0.13.0`.
>
> Our Terraform still provisions `compute_static` as a Fastly service dictionary. In the older SDK, `ConfigStore` was effectively backed by the dictionary hostcalls, so `ConfigStore::open("compute_static")` worked. In `0.13.0`, `ConfigStore` now opens a real Compute config store, so staging panicked because no config store named `compute_static` exists.
>
> The fix is to use `fastly::Dictionary::open("compute_static")` for this app, since that matches what Terraform actually creates. I added a comment and scoped `#[allow(deprecated)]` around that use so future SDK cleanups do not accidentally switch it back.
>
> Relevant links:
> - Fastly Rust SDK changelog: https://docs.rs/crate/fastly/0.13.0/source/CHANGELOG.md
> - Old `0.9.1` `ConfigStoreHandle` source using `fastly_dictionary::open`: https://docs.rs/crate/fastly/0.9.1/source/src/config_store/handle.rs
> - New `0.13.0` `ConfigStoreHandle` source using config-store APIs: https://docs.rs/crate/fastly/0.13.0/source/src/config_store/handle.rs
> - New `0.13.0` dictionary source: https://docs.rs/crate/fastly/0.13.0/source/src/dictionary/handle.rs


Let's try again

PR #1090, opened 2026-06-26
crates-io: Upgrade `aws` provider to v6

Turbo87:
This bumps the `aws` provider constraint for the crates-io module from `~> 5.64` to `~> 6.0`. It makes `cache_tag_config` available on CloudFront distributions (added in provider v6.46.0), which we would like to use in the future for improved cache invalidations.

v6 repurposes the `region` attribute on `aws_s3_bucket` for "Enhanced Region Support", so the bucket's actual region is now exposed as `bucket_region`. The three reads in the Fastly config are updated accordingly.

`fastly-tls-subscription` and `storage_transfer_iam` are shared with the `release-distribution` module, which stays on `~> 5.64`. We widen their constraints to `>= 5.64, < 7` (matching the ace-certificate pattern) so that both v5 and v6 callers resolve.

Lock files pin `aws` 6.52.0.

---

Disclaimer: I've used Claude Code to create this PR and I don't know enough about it to be 100% confident in the changes. Please apply and test these carefully.


terragrunt/modules/fastly-tls-subscription/_terraform.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5.64"
+      version = ">= 5.64, < 7"
```
marcoieni, line 7: `terragrunt/modules/release-distribution` also depends on this. 

> how do you know?

by running https://github.com/marcoieni/infratk with `graph --clipboard` and copying it into https://dreampuf.github.io/GraphvizOnline/ 😅
Same for `terragrunt/modules/storage_transfer_iam`.

I will update that module 👍

EDIT: maybe that module doesn't need to be updated because of the constraints you set 👍

marcoieni (approved):

terragrunt/modules/fastly-tls-subscription/_terraform.tf
Turbo87: > `fastly-tls-subscription` and `storage_transfer_iam` are shared with the `release-distribution` module, which stays on `~> 5.64`. We widen their constraints to `>= 5.64, < 7` (matching the ace-certificate pattern) so that both v5 and v6 callers resolve.

did you see this from the PR description? 😅

marcoieni: why I always skip pr descriptions when reviewing PRs? 😅😂

PR #1091, opened 2026-06-30
dev-desktop: Add lldb 20

aDotInTheVoid:
Maybe this will be new enough that we can run the `tests/debuginfo` on dev-desktops??

https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/dev-desktop.20debuginfo-lldb.20tests.20failing/with/607086157

### DIFF
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -59,6 +59,7 @@
       - flex # Allows running `x build gcc`
       - lld
       - lldb
+      - lldb-20
       - m4
       - mold
       - gdb
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -58,7 +58,6 @@
       - emacs
       - flex # Allows running `x build gcc`
       - lld
-      - lldb
       - lldb-20
       - m4
       - mold


ansible/roles/dev-desktop/tasks/dependencies.yml
```diff
@@ -59,6 +59,7 @@
       - flex # Allows running `x build gcc`
       - lld
       - lldb
+      - lldb-20
```
jieyouxu, line 61: Suggestion: I suspect we need at least lldb 21 for debuginfo tests, because currently in CI we only exercise lldb 21+ on aarch64 macos runners

aDotInTheVoid: Hmm, that will need to wait until dev-desktops are updated to 26.04: https://ubuntu.com/developers/docs/reference/availability/llvm/ (https://github.com/rust-lang/infra-team/issues/267)

Is the minimum lldb needed to test documented anywhere? (CC @walnut356)

jieyouxu: Likely not, since we're still figuring out the testing story.

Walnut356: Oh yeah that's why i ended up switching to fedora for gsoc 🫠 iirc the current tests should work ok for lldb 19+. I'm 100% certain they pass on 22 and 21. 20 and 19 are a bit more iffy, but i dont think anything major was added that the non-msvc tests rely on.

marcoieni:
let me know if this is ready to be reviewed or if the PR comment needs to be resolved

jieyouxu:
I think we can merge lldb-20 -- it's at least much newer than whatever was on dev desktops currently

marcoieni:
I think the packages `lldb` and `lldb-20` are not compatible between themselves.

Try it yourself:

```
docker run -it ubuntu:24.04 bash
apt update
apt install lldb
apt install lldb-20
```

output:
```
The following packages will be REMOVED:
  lldb lldb-18 python3-lldb-18
The following NEW packages will be installed:
  libclang-cpp20 liblldb-20 libllvm20 lldb-20 python3-lldb-20
```

is it ok uninstalling lldb-18 and having only lldb-20?
and also, the command `lldb` is no longer available, just `lldb-20`.

jieyouxu:
AFAIK not having lldb-18 is fine if we provide lldb-20

marcoieni:
ok, I removed the old lldb since lldb-20 removes it anyway.

PR #1095, opened 2026-07-01
remove allocation for const string

marcoieni:
It should be a trivial change. I removed the `format` which allocates.


ubiratansoares (approved):
Nice!

PR #1096, opened 2026-07-01
Reuse serialized request log

marcoieni:
## AI disclosure

I used GPT5.5 with the codex harness to generate this change. I reviewed its output.

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -3,7 +3,6 @@ use fastly::http::{header, Method, StatusCode, Version};
 use fastly::{Error, Request, Response};
 use log::{info, warn, LevelFilter};
 use log_fastly::Logger;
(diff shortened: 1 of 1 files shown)


terragrunt/modules/crates-io/compute-static/src/main.rs
```diff
@@ -285,15 +284,15 @@ fn build_and_send_log(log_line: LogLineV1Builder, config: &Config) {
...
         Ok(log) => {
             let versioned_log = LogLine::V1(log);
-
-            [
-                &config.datadog_request_logs_endpoint,
-                &config.s3_request_logs_endpoint,
-            ]
-            .iter()
-            .for_each(|endpoint| {
-                info!(target: endpoint, "{}", json!(versioned_log).to_string());
```
marcoieni, line 304: before we were serializing the same thing inside the loop

ubiratansoares (approved):
Small wins ftw!

PR #1097, opened 2026-07-01
Optimize download URL rewrite

marcoieni:
Remove the regex to save on Fastly compute duration. I assume code is faster written like this.

To test on mac: `cargo test --target aarch64-apple-darwin`
## AI disclosure

I used GPT5.5 with the codex harness to generate this change. I renamed a variable and added some comments.


marcoieni:
I deployed this on staging and it works: https://static.staging.crates.io/crates/KeyBoxen/0.2.0-pre.1/download
Fastly headers:
image

ubiratansoares (approved):

Turbo87:
👋  could you ping the crates.io for such PRs in the future, so that we are at least aware of such changes? 🙏

marcoieni:
yes, sorry!
did you notice any difference in dashboards or similar?

Turbo87:
> did you notice any difference in dashboards or similar?

no, all good AFAICT :)

PR #1098, opened 2026-07-01
fix datadog host in crates-io/compute-static

marcoieni:
Apparently this code wasn't tested on CI.
This code now uses the config.datadog_host variable, which was unused previously.
Tasks:

- [ ] can we add this crate to the workspace so that we don't need to edit CI?
- [ ] is using this variable correct here?

## AI disclosure

I used GPT5.5 with the codex harness to generate this change.


marcoieni:
solved by https://github.com/rust-lang/simpleinfra/pull/1122 and https://github.com/rust-lang/simpleinfra/pull/1121

PR #1102, opened 2026-07-04
dev-desktop: set vim alternative

joshtriplett:
The dev desktops install neovim, which accurately has a reputation for
having friendlier defaults. It's a good default for people who don't
have a specific preference for classic vim. So, use
`update-alternatives` to set `vim` to refer to neovim.

### DIFF
--- a/ansible/roles/dev-desktop/tasks/dependencies.yml
+++ b/ansible/roles/dev-desktop/tasks/dependencies.yml
@@ -85,6 +85,10 @@
       - libxcb-xfixes0-dev
     state: present
 
+# neovim is installed, and a good default for people without a specific preference for classic vim
+- name: Set vim alternative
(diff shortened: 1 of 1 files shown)


marcoieni:
I'm against this change. I don't think overriding ubuntu's default is a good idea. It might be surprising for developers discovering that vim links to neovim. 

People can set this themselves if they want for their user. 

I'm curious to see what others think.

joshtriplett:
> I don't think overriding ubuntu's default is a good idea.

Ubuntu, and Debian, have the same priority (30) for the /usr/bin/vim alternative provided by vim and neovim. AFAICT, they don't express a preference between the two. The alternative just points to whichever of vim and neovim is installed first.

```
/usr/bin/nvim - priority 30
/usr/bin/vim.basic - priority 30
```

PR #1106, opened 2026-07-08
crates-io: Wire up CDN cache tags for static.crates.io

Turbo87:
The crates.io backend attaches cache tags to the S3 objects of `static.crates.io` as comma-separated `cache-tags` metadata (`crate:{name}` and `release:{name}@{version}`). This PR makes both CDNs read that metadata at cache-fill time so that grouped, tag-based purges become possible instead of enumerating URL paths on deletion.

- **CloudFront:** `cache_tag_config` on the static distribution reads the `x-amz-meta-cache-tags` header and indexes cached objects under the tags, enabling `#crate:{name}` style invalidations. The attribute requires AWS provider >= 6.46.0, the lock files already pin 6.52.0.
- **Fastly:** the `compute-static` worker registers a `set_after_send` callback that translates the same header into surrogate keys before the response enters the cache.

r? @rust-lang/infra 

Note that this PR was created with the help of Claude Code and I had no way of verifying that it works on staging/production yet. This should be applied to the staging environment and then tested manually with tag-based invalidations.

### DIFF
--- a/terragrunt/modules/crates-io/cloudfront-static.tf
+++ b/terragrunt/modules/crates-io/cloudfront-static.tf
@@ -11,6 +11,19 @@ resource "aws_cloudfront_function" "static_router" {
   code    = file("${path.module}/cloudfront-functions/static-router.js")
 }
 
+resource "aws_cloudfront_response_headers_policy" "static" {
+  name = "${replace(var.static_domain_name, ".", "-")}--static"
+
+  // CloudFront reads the cache tags from the origin response at cache-fill time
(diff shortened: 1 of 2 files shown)


marcoieni:
Deployed in staging 👍
If you tell me it works as expected, I will deploy it in production as well.

Turbo87:
I've tested this on staging and it seems to work as intended. I've also added another commit to the branch to remove the cache-tags header from Cloudfront responses, to match how it is implemented for Fastly. The header is an implementation detail that we don't have to send to our clients.

PR #1119, opened 2026-07-13
import rate limit configured via the dashboard

marcoieni:


marcoieni:
we decided to not put this in terraform and leave it configurable from the ui

PR #1122, opened 2026-07-15
crates.io: Test the static.crates.io Fastly worker

Turbo87:
Cargo now runs the `static.crates.io` worker’s `wasm32-wasip1` test binaries through Viceroy, which supplies Fastly hostcalls unavailable in a regular WASI runtime. 

The worker README documents the local `cargo nextest run` workflow and explains why each test runs in an isolated WebAssembly instance. 

CI exercises the same test suite in a dedicated job with pinned Viceroy and cargo-nextest versions backed by job-local Rust caching. Renovate’s GitHub Actions versions preset keeps both tool pins current.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,6 +1,9 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["github>rust-lang/renovate"],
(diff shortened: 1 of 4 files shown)


marcoieni (approved):
thanks!

PR #1120, opened 2026-07-15
crates.io: Enable dynamic compression for text responses

Turbo87:
This sets [`X-Compress-Hint`](https://www.fastly.com/documentation/reference/http/http-headers/X-Compress-Hint/) and `Vary: Accept-Encoding` on eligible text, JSON, and XML responses to enable response compression for these files.

PR #1127, opened 2026-07-15
crates.io: Enable compression for index.crates.io responses

Turbo87:
This marks complete successful GET responses for Fastly dynamic compression while keeping cached objects uncompressed.

It adds `Accept-Encoding` to `Vary` for downstream caches.

### Related

- #1120


marcoieni:
applied to staging 👍

Turbo87:
after testing this we've realized that the segmented caching on Fastly for index.crates.io is preventing this from working correctly. this will need a different approach unfortunately. it is still tbd what the best approach is, so I'll close this PR for now.

PR #1129, opened 2026-07-17
docs(terraform): clarify queue visibility timeout comment

syphar:


marcoieni:
I'm closing this since on zulip we agreed that we can just use the aws default value

PR #1130, opened 2026-07-18
fix: include Content-Length in fastly HEAD responses

lf-:
I hit an extremely strange issue while working on buck2: I got an error implying I needed to explicitly specify a content length while downloading crates at content-addressed paths. I dug through the code, realized there was a fallback path of making a HEAD request to get the length, then pulled out `curl` and went "wtf".

This only happens on fastly:

```
$ curl --head 'https://fastly-static.crates.io/crates/adler2/2.0.1/download' | grep -i content-
content-type: application/gzip

$ curl --head 'https://cloudfront-static.crates.io/crates/adler2/adler2-2.0.1.crate' | grep -i content-
content-type: application/gzip
content-length: 13366
```

With a GET request of an empty range, the content length is correct:

```
$ curl -sS -D - -o /dev/null --http1.1 -H 'Range: bytes=0-0' 'https://static.crates.io/crates/adler2/2.0.1/download' | grep -i content-
Content-Length: 13366
content-type: application/gzip
```

Validation: I'm terribly sorry; I don't know how to add automated tests to this project that would actually test this. It seems like it's not compatible with viceroy, due to `Request::set_after_send` not being supported; this implies to me that the test infra necessary to write an automated test doesn't exist.

For what it's worth, I have thrown an agent at making throwaway garbage LLM harness code that seems to validate this change *does* cause it to start sending Content-Length on HEAD requests where it did not before (I will spare you that code. it's not good). It cfg'd out calling `set_surrogate_keys` which apparently was pretty much all that was needed to get this crate to work on Viceroy, fyi, in case you *do* want to build a test suite.

### DIFF
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -298,6 +298,16 @@ fn send_request_to_s3(config: &Config, request: &Request) -> Result<Response, Er
 
     enable_dynamic_compression(request, &mut response);
 
+    // Without this, HEAD requests throw away the content length from the upstream S3.
+    // The default Automatic mode will re-frame the response depending on
+    // client preferences, which may differ from the way content is fetched
+    // from the origin. HEAD doesn't *do* framing since it has no body, and for
+    // some reason, the automatic mode doesn't attach any content-length to the
+    // response.
+    if request.get_method() == Method::HEAD {
+        response.set_framing_headers_mode(fastly::http::FramingHeadersMode::ManuallyFromHeaders);
+    }
+
     Ok(response)
 }
 
--- a/terragrunt/modules/crates-io/compute-static/src/main.rs
+++ b/terragrunt/modules/crates-io/compute-static/src/main.rs
@@ -302,7 +302,7 @@ fn send_request_to_s3(config: &Config, request: &Request) -> Result<Response, Er
     // discards S3's Content-Length. This is normally necessary because downstream
     // transformations such as client-negotiated compression can change the
     // representation. For HEAD, preserve the origin length only when the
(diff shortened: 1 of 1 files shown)


lf-:
hm. regarding your rewrite here; i suspect it's possible to, for head requests specifically, to refuse to compress them for the client (since.. surely dynamic compression doesn't actually know the size if it's just sending the request through to the origin?) and return the uncompressed size. i actually wonder how s3 itself implements this, because it's subtle and would change the algorithmic complexity of the HEAD method a lot.

marcoieni:
ok, the PR is ready to be reviewed now (finally finished my back and forth with AI) 😅
Wdyt? I will deploy to staging now

marcoieni:
deployed to staging

lf-:
I think this doesn't fix the bug: if clients don't set a range, then Fastly will continue discarding the content length as is the case today, which is unfortunate. I guess I'm not sure what dynamic compression achieves for HEAD specifically: it *seems* as if the thing that you'd want to do is to say "no i didn't compress it" and always give the actual length for HEAD specifically. I'm curious what other implementations do though.

marcoieni:
> if clients don't set a range, then Fastly will continue discarding the content length as is the case today, which is unfortunate.

the error you reported in your first post is now fixed in staging, no? 🤔

```
$ curl --head \
  https://fastly-static.staging.crates.io/crates/crossbeam/0.2.10/download \
  | grep -i content-
content-type: application/gzip
content-length: 32996
```

lf-:
Testing confirms that crossbeam-0.2.10 fetches as expected with buck2, so yeah, I think this is fixed. Thanks! I didn't actually know how to use the staging endpoint.

PR #1136, opened 2026-07-23
create safe-read-only datadog role

marcoieni:


terraform/team-members-datadog/safe-readonly.tf
```diff
@@ -0,0 +1,25 @@
+# Opt out of all default permissions and grant
+# only the observability access needed to debug incidents and troubleshoot issues.
+# This role can be used by AI agents to access DataDog.
```
Turbo87, line 3: keep in mind that those AI agents would potentially still have access to PII and secrets that leak into the logs. definitely better than read-write access, but still not entirely "safe".

marcoieni: Yes, ideally we don't have PII and secrets in the logs. Maybe we can use this role to detect them 😎
Anyway, I will document this concern

Turbo87: we're handling IP addresses, we will always have PII like that in the logs, otherwise the logs would be pretty useless 😅

Issue #1132, opened 2026-07-18
Evaluate replacing CodeBuild with custom self-hosted runners


Mark-Simulacrum:
In general it looks like rolling our own self-hosted runner integration that launches an EC2 instance just-in-time with an ephemeral GHA runner should be fairly straightforward and would allow reducing our credit usage 3x at on-demand prices to 8x for spot prices.

The security should be basically the same: we would run on a fresh EC2 instance per-run, launched from latest Ubuntu (or w/e base) AMI. That instance wouldn't have any inbound ports open in its security group and would run with minimal credentials (maybe none, maybe just DataDog agent, TBD). The instance would get automatically killed after X hours of runtime.

| Build location | Type | Duration | Price on-demand | Price spot |
|-|-|-|-|-|
| CodeBuild | dist-quick | 2h | 9.576 | n/a |
| c8a.16x | dist-quick | 1h | 3.4486 | 1.20 |
| CodeBuild | dist | 3h | 14.364 | n/a |
| c8a.16x | dist | 1h 11m | 3.4486 | 1.20 |

With parallel-rustc:

* c8a.12x full dist 1h 19m 35s - https://github.com/rust-lang/rust/actions/runs/29787213997
* c8a.12x dist-quick 55m - https://github.com/rust-lang/rust/actions/runs/29791306969

Without paralllel rustc:

* c8a.12x full dist 1h 24m 7s - https://github.com/rust-lang/rust/actions/runs/29829979202/job/88632263692
* c8a.12x quick dist 59m - https://github.com/rust-lang/rust/actions/runs/29918517367

Without parallel, without /tmp tmpfs (custom AMI):
* c8a.12x quick 56m - https://github.com/rust-lang/rust/actions/runs/31064471255/job/92499270219
* c8a.8x quick 59m - https://github.com/rust-lang/rust/actions/runs/31099033462/job/92608007797
* c8a.4x quick 1h10m - https://github.com/rust-lang/rust/actions/runs/31103399972/job/92622541550
* c8a.2x quick failed - https://github.com/rust-lang/rust/actions/runs/31130092566/job/92717055770 (unclear why)
* m8a.2x quick 1h 35m - https://github.com/rust-lang/rust/actions/runs/31142656720/job/92755682675
* m8azn.6x quick 57m - https://github.com/rust-lang/rust/actions/runs/31232686921/job/93039515563
[...]

PR #1141, opened 2026-07-25
Add self-hosted runner configuration

Mark-Simulacrum:
This adds a role in the CI accounts that bors can assume, which grants permission to describe instances, terminate instances, and run instances. It also allows resolving Ubuntu AMI IDs.

The instance launch is restricted to using a launch template which constraints some of the options available to bors. We might remove parts of that enforcement to make it easier to experiment with different CI configurations.

Depending on frequency of changes to the resources defined here, we may want to expose versions or test against the staging account from production (maybe limited to a subset of try builds). Otherwise once we apply terraform updates (infra-admin only) the next or ongoing build gets impacted immediately which is not amazing.

Right now the assumption made here is that bors will launch instances and then at boot initialize them via a user-data script to configure/install the dependencies needed (GitHub actions runner, docker, etc.). I'm still exploring our options for building an AMI instead -- my current sense is that if we do build AMIs it should probably be in rust-lang/rust CI and treated similar to docker (i.e., we rebuild on every use of this runner as a pre-step before launching it, caching against a ~hash of the inputs, which would include the base AMI ID). But I'm still not sure building AMIs actually buys us enough value to be worth it since the installation steps are pretty fast anyway.

cc https://github.com/rust-lang/simpleinfra/issues/1132

cc @Kobzol

### DIFF
--- a/terragrunt/modules/ci-runners/instance.tf
+++ b/terragrunt/modules/ci-runners/instance.tf
@@ -1,0 +1,45 @@
+resource "aws_launch_template" "runner" {
+  name = "gha-runner"
+
+  block_device_mappings {
+    device_name = "/dev/sda1"
+    ebs {
+      delete_on_termination = true
+      volume_type           = "gp3"
+      volume_size           = 300
+      throughput            = 800
+      iops                  = 3500
+    }
+  }
+
+  ebs_optimized = true
+
+  metadata_options {
+    http_endpoint               = "enabled"
+    http_tokens                 = "required"
+    http_put_response_hop_limit = 1
+    instance_metadata_tags      = "disabled"
+  }
+
+  network_interfaces {
+    // We use a public IP address rather than NAT gateways because it's both
+    // easier and cheaper. Each NAT gateway is a minimum of $32/month,
+    // whereas a public IPv4 IP is $3.6/month. And we typically want 3 NAT
+    // gateways (one per AZ) to avoid cross-AZ bandwidth charges. We also don't
+    // expect to always have live runner instances.
+    //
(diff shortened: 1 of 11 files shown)


terragrunt/modules/ci-runners/instance.tf
```diff
@@ -0,0 +1,42 @@
+resource "aws_launch_template" "runner" {
+  name = "gha-runner"
+
+  block_device_mappings {
```
Kobzol, line 4: Can I use this template when launching the EC2 instance to avoid hardcoding the JSON with the disk parameters in the "start EC2 instance" command?

Kobzol (commented):
Thanks! I don't understand most of this, but still took a look :)

terragrunt/modules/ci-runners/instance.tf
Mark-Simulacrum: Yes, in fact, you are required to by the permissions given here. This is the configuration that works locally for me (though I've not tested all the way through -- I suspect the user data script will need altering, the removal of IMDSv2 means the instance can't self-discover e.g. its region, we may want to not give it credentials while still enabling the metadata interface):

```
    let output = aws()
        .arg("ec2")
        .arg("run-instances")
        .arg("--launch-template")
        .arg("LaunchTemplateName=gha-runner,Version=$Latest")
        .arg("--image-id")
        .arg(format!("resolve:ssm:{parameter}"))
        .arg("--instance-type")
        .arg(instance_type)
        .arg("--instance-market-options")
        .arg("MarketType=spot")
        .arg("--user-data")
        .arg(script)
        .output()?;
```

PR #1149, opened 2026-08-05
configure fastly NgWaf for docs.rs, prepare integration with WASM module

syphar:


terraform/docs-rs/fastly.tf
```diff
@@ -32,6 +32,10 @@ resource "fastly_configstore_entries" "docs_rs" {
     # max age for HSTS header.
     # should be less for test / staging environments
     hsts_max_age = "31557600"
+
+    # TODO: switch on when product_enablement went through?
+    # ngwaf_corp      = "some_corp" # FIXME: what's our corp name?
+    # ngwaf_workspace = fastly_ngwaf_workspace.webapp.id
```
syphar: @marcoieni I'm actually not sure if this is correct right now. 

Workspace id is probably <> workspace name, so I could imagine when using the workspace id, I don't need the corp name. 

But in the fastly rust library I don't see where to pass in the workspace id instead of the name.

marcoieni:
https://manage.fastly.com/security/ngwaf/workspaces/nAq6HXr6bLdWNT3wScMFYE
image

terraform/docs-rs/fastly.tf
marcoieni:
```suggestion
    # ngwaf_workspace = fastly_ngwaf_workspace.webapp.name
```
yes, passing the name should work 👍

terraform/docs-rs/fastly.tf
```diff
@@ -32,6 +32,10 @@ resource "fastly_configstore_entries" "docs_rs" {
     # max age for HSTS header.
     # should be less for test / staging environments
     hsts_max_age = "31557600"
+
+    # TODO: switch on when product_enablement went through?
+    # ngwaf_corp      = "some_corp" # FIXME: what's our corp name?
```
marcoieni:
```suggestion
    # ngwaf_corp      = "rust_foundation"
```
best corp ever of course 😎

syphar:
@marcoieni how would we continue here? 

I think I should get a user / permissions to play around before we switch it on? 

Since this is prod, we probably go in steps? or can we directly set corp/workspace and likely we're fine? (I think so)

marcoieni:
> I think I should get a user / permissions to play around before we switch it on?

Done! I did it from the UI, it's not possible to give from terraform. See https://github.com/fastly/terraform-provider-fastly/issues/1209

marcoieni (approved):

marcoieni:
applied!

PR #1152, opened 2026-08-05
remove monitoring pagerduty integration

marcoieni:
This alert is already in DataDog. We are switching to Datadog on call.

- [x] apply
- [x] remove the `alertmanager-receiver-pagerduty-cratesio` SSM parameter


marcoieni:
the [new monitor](https://app.datadoghq.com/monitors/311304803) in datadog works: 
image

PR #1154, opened 2026-08-06
provision docs-rs fastly waf

marcoieni:
Extracts just the terraform part from the PR https://github.com/rust-lang/simpleinfra/pull/1149
This PR just provisions the NgWaf without configuring it.
I extracted it into a separate PR because I also had to upgrade the Fastly provider.


terraform/docs-rs/fastly.tf
```diff
@@ -60,6 +60,16 @@ resource "fastly_service_compute" "docs_rs" {
     port    = 80
   }
 
+  product_enablement {
+    domain_inspector = true
```
marcoieni, line 64: I found it already enabled, and so I left it here.

marcoieni:
I already applied this. but I can't merge due to github actions outage.

syphar (approved):

PR #1165, opened 2026-08-13
connect docs.rs fastly CDN with NgWaf

syphar:


syphar:
r? @marcoieni 

let's apply this when both of us are available, though nothing should break.

marcoieni (approved):

marcoieni:
reverted this because Denis told me there were issues.

PR #1167, opened 2026-08-13
allow docs.rs fastly compute service for NgWaf edge deployment

syphar:


syphar:
right now we think we don't need this.

PR #1185, opened 2026-08-20
crates-io: explain 429 error

marcoieni:
try making cases like [https://github.com/rust-lang/crates-io-auth-action retries](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/https.3A.2F.2Fgithub.2Ecom.2Frust-lang.2Fcrates-io-auth-action.20retries/with/617527789) less confusing

### DIFF
--- a/terragrunt/modules/crates-io/fastly-webapp.tf
+++ b/terragrunt/modules/crates-io/fastly-webapp.tf
@@ -1,7 +1,8 @@
 // This file configures the crates.io web app API via Fastly CDN
 
 locals {
-  fastly_webapp_domain_name = "fastly-app.${var.webapp_domain_name}"
+  fastly_webapp_domain_name       = "fastly-app.${var.webapp_domain_name}"
+  webapp_rate_limit_response_code = 429
 }
 
 resource "fastly_service_vcl" "webapp" {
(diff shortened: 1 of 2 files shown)


Turbo87 (commented):
since this is for API requests it should ideally follow the JSON error convention that we use on the backend, so that downstream tools can display errors consistently

terragrunt/modules/crates-io/fastly-webapp.tf
```diff
@@ -92,6 +93,19 @@ resource "fastly_service_vcl" "webapp" {
...
   }
 
+  # Add a response body to requests blocked by the NGWAF rate limit.
+  snippet {
+    name    = "NGWAF rate limit response body"
+    type    = "error"
+    content = <<-VCL
+      if (obj.status == ${local.webapp_rate_limit_response_code}) {
+        set obj.http.Content-Type = "application/json";
+        synthetic {json"{"errors":[{"detail":"too many requests from the same IP"}]}"json};
```
marcoieni, line 103: the json syntax is documented in https://www.fastly.com/documentation/reference/vcl/types/string/

PR #1195, opened 2026-08-26
remove Paul from team members

marcoieni:


marcoieni:
I applied this PR, but I cannot merge it because there is a GitHub incident.

PR #1196, opened 2026-08-26
docs.rs CDN: set up tracing & configure datadog logging for application logs

syphar:
I think this should be it. 

Related links: 
- https://www.fastly.com/documentation/guides/integrations/streaming-logs/setting-up-remote-log-streaming-for-compute/

Generally: 
This just uses the standard tracing json log format, depending on what we want to do in datadog we either need to pre-process these logs in datadog, or rename fields in here. 

I'm not sure about the api-key / secret thing, feel free to tell me what should be different. 

Also: **these are just error / info logs from the wasm module itself**. We don't have any access log yet. 

Future improvements: 
- start adding request-id, create a surrounding span that includes it so we can see it in the logs.


terraform/docs-rs/fastly-compute-docs-rs/src/logging.rs
```diff
@@ -0,0 +1,31 @@
...
+
+// Must match the logging endpoint name in Terraform.
+const APPLICATION_LOG_ENDPOINT: &str = "application_logs";
+
+pub(crate) fn setup() {
+    let application_logs = fastly::log::Endpoint::from_name(APPLICATION_LOG_ENDPOINT);
+
+    tracing_subscriber::registry()
+        .with(
+            // print to stdout for fastly log-tail. With compact format.
```
marcoieni: I think you use stderr, right?

syphar: right!

terraform/docs-rs/fastly.tf
```diff
@@ -3,6 +3,10 @@ locals {
   config_store_name = "docs_rs_config"
 }
 
+data "aws_ssm_parameter" "datadog_api_key" {
+  name = "/prod/docs-rs/datadog/api-key"
```
marcoieni:
```suggestion
  name = "/prod/docs-rs/datadog-api-key"
```
more similar to our conventions

syphar:
Some research showed we're missing some fields: 

- https://docs.datadoghq.com/api/latest/logs/send-logs/?code-lang=rust
- https://www.fastly.com/documentation/guides/integrations/logging-endpoints/log-management-and-observability/log-streaming-datadog/#recommended-log-format

will improve this and re-request a review, perhaps include the request id.

syphar:
So, "small" update here. 

For now I'm using the help of the `tracing-serde` crate to build my own tracing layer, and a custom struct for the log format that datadog expects. 

Generally I would like to extract that stuff into a separate small crate (`tracing-fastly` or so), so we (and others) can reuse it.

Cargo.toml
```diff
@@ -5,11 +5,7 @@ resolver = "2"
 members = [
     "setup-deploy-keys",
     "ansible/roles/dev-desktop/files/team_login",
+]
+exclude = [
```
syphar, line 9: that's because our test-suite now only runs with the fastly test-runner, not as normal tests any more.

syphar:
so the main reason for all the code in `crate::logging` is 
- provide the datadog json format 
- collect fields from tracing spans, also emit them on the log.

.github/workflows/main.yml
```diff
@@ -23,30 +23,40 @@ jobs:
...
       VICEROY_VERSION: "0.20.0"
       # renovate: datasource=crate depName=cargo-nextest
       NEXTEST_VERSION: "0.9.140"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: docs.rs
+            workspace: terraform/docs-rs/fastly-compute-docs-rs
+            test-command: cargo install just --locked && just lint run-unit-tests
```
marcoieni: you can install `just` from apt.
I think `cargo install` is very slow when there's a cache miss.

syphar: valid, will try

Cargo.lock
```diff
@@ -413,6 +413,11 @@ name = "fastly-compute-project"
 version = "0.1.0"
 dependencies = [
  "fastly",
+ "serde",
+ "serde_json",
+ "tracing",
+ "tracing-serde",
```
marcoieni: if tracing-serde is used by `terraform/docs-rs/fastly-compute-docs-rs`, which is excluded from the workspace. Why is this here? 🤔

Cargo.toml
marcoieni: do we need to explicitly exclude the package? If yes, should we also exclude `terragrunt/modules/crates-io/compute-static`?

terraform/docs-rs/fastly-compute-docs-rs/Cargo.toml
```diff
@@ -6,5 +6,15 @@ publish = false
...
 uuid = { version = "1.25.0", features = ["v4", "fast-rng"] }
+tracing = "0.1"
+tracing-serde = "0.2"
+tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt", "json"] }
 
+[profile.fastly-release]
+inherits = "release"
+debug = 1
+codegen-units = 1
+lto = "fat"
```
marcoieni, line 20: should we also add this to `terragrunt/modules/crates-io/compute-static`?
Since it's not in the workspace anymore

terraform/docs-rs/fastly.tf
```diff
@@ -44,6 +48,12 @@ resource "fastly_configstore_entries" "docs_rs" {
 resource "fastly_service_compute" "docs_rs" {
   name = local.domain_name
 
+  logging_datadog {
+    name   = "application_logs"
```
marcoieni, line 52: this is very generic. Is it how it will show in the dashboard? Should we call it `docs-rs-cdn` or similar?

syphar: that's just the name of the endpoint in the fastly compute service. 

How it's called in datadog depends on the `service_name` I'm providing

terraform/docs-rs/fastly-compute-docs-rs/Cargo.toml
syphar: hm... that's interesting. from what I see, `crates-io/compute-static` wasn't in the workspace either, so the profile didn't exist there. 

But also, looking at the `fastly.toml` for compute-static, that profile isn't referenced. 

So IMO, it's an unrelated fix / change.

Cargo.lock
syphar: probably historical from my back and forth trying to fix stuff. Will try to revert and see if it gets updated again.

syphar: I updated it

marcoieni:
ok, I will apply this now. You can squash your commits if you want, otherwise I can merge like this.

Cargo.toml
syphar: it seems like our package needs to be excluded, but the compute-static one not. didn't dig into why yet.

PR #1197, opened 2026-08-26
docs-rs CDN: generate request ID in CDN

syphar:
will be useful for application and access logs later. 

Also we will see it on our origin servers.

### DIFF
--- a/terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
+++ b/terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
@@ -9,6 +9,7 @@ use fastly::{
         header::{CACHE_CONTROL, EXPIRES, STRICT_TRANSPORT_SECURITY},
     },
 };
+use uuid::Uuid;
 
 // Should match the backend name in terraform
 const DOCS_RS_BACKEND: &str = "docs_rs_origin";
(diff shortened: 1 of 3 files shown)


terraform/docs-rs/fastly-compute-docs-rs/src/main.rs
```diff
@@ -82,6 +84,14 @@ fn main(mut req: Request) -> Result<Response, Error> {
         }
     }
 
+    // Generate the request ID at the POP receiving the client request. On
+    // shield POPs, preserve the ID set by the edge POP.
+    if shield.response_is_for_client() {
+        req.set_header(X_REQUEST_ID, Uuid::new_v4().to_string());
```
marcoieni, line 90: did you consider using https://docs.rs/fastly/latest/fastly/struct.Request.html#method.get_client_request_id ?
```suggestion
        let request_id = req
            .get_client_request_id()
            .context("this is the client request, it should have a request ID")?
            .to_owned();
        req.set_header(X_REQUEST_ID, request_id);
```

Disclaimer: finding found by AI

syphar: I think that's not what we want: 
https://www.fastly.com/documentation/guides/compute/developer-guides/sandbox-lifecycle/#per-request-id

marcoieni (approved):
applied

PR #1216, opened 2026-09-02
update dev-desktops terraform providers

marcoieni:
plan was clean


terraform/dev-desktops/regions.tf
```diff
@@ -13,7 +13,7 @@ module "aws_eu_central_1" {
     "dev-desktop-eu-1" = {
       instance_type = "c6g.8xlarge"
       instance_arch = "arm64"
-      storage       = 3000
+      storage       = 6000
```
marcoieni, line 16: these were manually changed with the UI.

PR #1220, opened 2026-09-02
bump playground instance type and size

marcoieni:
double cores and RAM and move to c7a to increases single core performance.
The playground should be faster.

c8a isn't present in the playground region.


shepmaster:
I have no objections to this change.

I haven't gotten any feedback that the playground is slow, so I don't know how much this will impact people in the real world. If you are happy to pay whatever the increased cost is, go for it!

PR #1221, opened 2026-09-02
disable crates-io webapp cloudfront distribution

marcoieni:
People can go directly to `cloudfront-app.crates.io` bypassing Fastly NGWAF, therefore rate limits. We prevent this by disabling the cloudfront CDN for the crates-io app.

### DIFF
--- a/terragrunt/accounts/legacy/crates-io-prod/crates-io/terragrunt.hcl
+++ b/terragrunt/accounts/legacy/crates-io-prod/crates-io/terragrunt.hcl
@@ -34,6 +34,8 @@ inputs = {
   index_cloudfront_weight = 1
   index_fastly_weight = 255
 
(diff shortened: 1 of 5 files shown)


Turbo87:
Instead of dropping it completely, could we:

- keep it for the staging environment, so that we can e.g. test cloudfront caching behavior reliably?
- keep it for the production environment, but require a secret header or something like that?

marcoieni:
I don't get why you need to test cloudfront since we don't use it at all at the moment for the webapp 🤔

Turbo87:
at the moment, yes, but if we keep it as a fallback then we should have some way to verify that it still works as expected if we ever had to use the fallback

marcoieni:
> keep it for the production environment, but require a secret header or something like that?

I added the variable `webapp_cloudfront_enabled` so that it is easy to enable cloudfront again in production in case we need it 👍

PR #1222, opened 2026-09-02
Increase CPU allocation for triagebot to one full CPU

Kobzol:
Apparently, we should be spending more credits :laughing: So why not give a whopping full CPU to triagebot, instead of a quarter (!) of a CPU that it had previously.

CC @rust-lang/triagebot

### DIFF
--- a/terraform/shared/services/triagebot/main.tf
+++ b/terraform/shared/services/triagebot/main.tf
@@ -44,8 +44,8 @@ module "ecs_task" {
   source = "../../modules/ecs-task"
 
   name   = "triagebot"
-  cpu    = 256
-  memory = 512
+  cpu    = 1024
+  memory = 2048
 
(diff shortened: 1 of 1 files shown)


marcoieni (approved):

apiraino:
Are the additional resources needed?
One way to keep in check an application bloat (or looking at what is taking up resources) is seeing how it behaves under contrained resources.

marcoieni:
I'll apply once you have an agreement in the team

Kobzol:
> Are the additional resources needed? One way to keep in check an application bloat (or looking at what is taking up resources) is seeing how it behaves under contrained resources.

Having more RAM is useful to cache more entries in our GitHub viewers. And generally increasing resources should lessen latencies a bit.

marcoieni (approved):
applied

PR #1235, opened 2026-09-15
crates.io: Resolve static asset URLs at the Fastly edge

Turbo87:
Matching `GET` and `HEAD` requests for crate downloads and READMEs now resolve to `static.crates.io` without contacting Heroku. Fastly returns a redirect by default and a JSON URL when `Accept` requests JSON.

---

Disclaimer: This was written by Codex (GPT-6 Astra), and due to missing infra access I could not verify if this actually works as intended.


ubiratansoares:
`terragrunt plan` looks good, I'll deploy these changes to `staging` first and give them a manual try.

Turbo87:
> I'll deploy these changes to `staging` first and give them a manual try

I just ran a quick manual test and things seems to be working as intended on staging, while production still routes to Heroku for now.

terragrunt/modules/crates-io/vcl/static-redirect-recv.vcl
```diff
@@ -0,0 +1,17 @@
...
+  declare local var.version STRING;
+  declare local var.resource STRING;
+  set var.name = re.group.1;
+  set var.version = re.group.2;
+  set var.resource = re.group.3;
+
+  set var.version = regsuball(var.version, "[+]", {"%2B"});
+  if (var.resource == "download") {
+    error 600 "/crates/" var.name "/" var.name "-" var.version ".crate";
+  } else if (req.http.User-Agent && req.http.User-Agent != "Amazon CloudFront") {
```
ubiratansoares: If I run

```
curl -v --max-redirs 0 -H "User-Agent;" https://fastly-app.staging.crates.io/api/v1/crates/crossbeam/0.2.10/readme
```

I get 302, but I was wondering whether we should validate the empty  UA _like_ Heroku in this case. The fix would be validating UA length.

Wdyt?

terragrunt/modules/crates-io/vcl/static-redirect-error.vcl.tftpl
```diff
@@ -0,0 +1,23 @@
...
+  set obj.http.Access-Control-Allow-Origin = "*";
+  set obj.http.Strict-Transport-Security = "max-age=31536000";
+
+  std.collect(req.http.Accept);
+  if (req.http.Accept ~ "json") {
+    set obj.status = 200;
+    set obj.response = "OK";
+    set obj.http.Content-Type = "application/json";
+    set obj.http.X-Frame-Options = "SAMEORIGIN";
+    set obj.http.X-XSS-Protection = "0";
```
ubiratansoares: Claude suggested adding

```
set obj.http.X-Content-Type-Options = "nosniff";
```

so we mirror what we have [in the backend](https://github.com/rust-lang/crates.io/blob/main/src/middleware/common_headers.rs#L48)

ubiratansoares (commented):
Found two things, partially assisted by Claude Code

terragrunt/modules/crates-io/vcl/static-redirect-error.vcl.tftpl
Turbo87: that prevents browsers from ignoring the `Content-Type: application/json` header, but since the content is still regular JSON, it can't really be misinterpreted as JS, CSS or HTML.

I've also removed the unnecessary `X-Frame-Options` and `X-XSS-Protection` headers. Those are not really relevant in this case either. I thought I had removed them earlier, but apparently they snuck back in during a rebase 🙈

terragrunt/modules/crates-io/vcl/static-redirect-recv.vcl
Turbo87: yeah, the user-agent filtering was completely unnecessary. I've removed it.

ubiratansoares (approved):
Applied this to `prod`.

